Issue #41, opened 2014-12-04
Set up Vagrant box w/ full production environment for development


ethanal:
Also update fabric.py to remove references to sandboxes

Issue #17, opened 2014-08-15
Remove ckeditor branding/styling or find a better WYSIWYG editor


ethanal:

ethanal:
ckeditor is fine. I found a different skin that looks a little better.

PR #43, opened 2015-01-02
Updating docs to reflect what worked today

theycallmezeal:
Following the setup instructions word-for-word didn't work out, so here's what did.

If I haven't introduced myself, Ethan: hi, call me Adam or Zeal, I don't care, I'm a junior, not a sysadmin, have hung out in CSL since Fox dragged me there in 2012 and will try to contribute more than a four-line diff once it's not 1:29 AM any more.

### DIFF
--- a/docs/setup/sandbox.rst
+++ b/docs/setup/sandbox.rst
@@ -6,7 +6,7 @@ Setting up Your Sandbox
 Virtualenv
 ----------
 
-SSH into Ion.
+SSH into Ion. You will need a CSL principal and to be added to the ion group.
 
 Navigate to ``/usr/local/virtualenvs``:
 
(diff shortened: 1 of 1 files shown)


ethanal:
Hi Adam - thanks for the PR! Between now and the release, we're going to need more than just James and I working on finishing it up so it's great that you want to contribute.

As for setting up sandboxes, we need to work on getting #41 closed. @ahamilto said a while ago that he would help out getting Gentoo to work in a Vagrant box and getting the local domain working over VPN. For now, though, working on the ion server itself is fine, but be mindful that you may run into issues with SQLite (hence the need for a dev environment matching the production environment). SQLite doesn't handle nulls and empties very well.

If you have time and want to start figuring out how everything is organized, take a look at issues #17 and #8.

PR #335, opened 2015-11-27
Master

jacobajit:
Added service registration script, called from the login page
Service worker doesn’t do anything right now, but satisfies the
conditions to show Chrome web app install banner
In the future, will be used to send push notifications and cache for
offline use/faster loading


pefoley2:
Minor nits, but it's helpful if you put a description of the pull request in the title.
You might also want to consider using git rebase to cleanup your history.

PR #339, opened 2015-12-01
Master

jacobajit:
Fix service worker path


jwoglom:
let's see if it works.

PR #345, opened 2015-12-04
Minor Fixes

tcyrus:
Some Accessibility Fixes and Template Fixes

### DIFF
--- a/intranet/apps/auth/forms.py
+++ b/intranet/apps/auth/forms.py
@@ -20,8 +20,8 @@ class AuthenticateForm(AuthenticationForm):
             The password text field.
 
     """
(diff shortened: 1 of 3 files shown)


pefoley2:
What is aria-label?

tcyrus:
https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/ARIA_Techniques/Using_the_aria-label_attribute

pefoley2:
You should note that somewhere then.
Is this the only usage? Should it be added elsewhere?
Is there something specific about this element that requires it?

jwoglom:
Aria-label is a standardized accessibility tag; nothing wrong with having it there

pefoley2:
@jwoglom I'm not saying it's wrong, just raising some other possible concerns :)

PR #349, opened 2015-12-08
Added View Roster / Close Roster Toggle

tcyrus:


jwoglom:
So this adds a "close roster" button after viewing a roster, in the "view roster" button's place? (Next to sign up?)

tcyrus:
It changes the text of the View Roster button, it does not replace the button.

PR #362, opened 2016-01-09
changed 'often signups' button and ldap popup shows up only once

2019okulkarn:


jwoglom:
don't merge until issue corrected

PR #371, opened 2016-01-24
move logo below emergency dialog

2019okulkarn:


pefoley2:
Do we need a totally different css file? looks like we could factor some stuff out to reduce duplication.

2019okulkarn:
@pefoley2 the new css file is needed because the position of the logo can't be set relative to the emergency dialog. It would look like this if there was no dialog box and the top margin was the same: http://puu.sh/mHXz6/c2d861068b.jpg.

pefoley2:
What i'm saying it it looks like there's stuff common to both login.css and emerg.css, so it might make sense to factor out the stuff that changes into separate files, and have a login_common.css with the stuff that remains the same.

2019okulkarn:
Do you mean choosing whether to @import login.css or emerg.css?

pefoley2:
I'm saying factor out the common stuff into a separate file and import that unconditionally.

jwoglom:
Don't use an @import statement; http://www.stevesouders.com/blog/2009/04/09/dont-use-import/

jwoglom:
Have everything in login_common.css moved back to login.css, then have emerg.css override whatever styles are necessary.

PR #397, opened 2016-02-26
Added feature to download folders as zip files

ezwang:
The interface isn't pretty, it could be changed to a right click menu or something
The max file size limit is 200MB of uncompressed data
It would be better to zip the files remotely before copying them over, but that would be harder to implement (differences between windows, linux)


pefoley2:
Mind writing a test?
Otherwise, looks good.

jwoglom:
Add tests in a future pull request if you can, please. Going to merge for now to test.

Issue #387, opened 2016-02-19
Blacklist for activity


jwoglom:
Students who can't sign up

PR #400, opened 2016-02-27
added activity blacklisting

ezwang:
users will not be able to sign up for an activity if they are on the activity's blacklist
fixes issue #387

### DIFF
--- a/intranet/apps/eighth/forms/admin/activities.py
+++ b/intranet/apps/eighth/forms/admin/activities.py
@@ -90,7 +90,7 @@ class ActivityForm(forms.ModelForm):
     def __init__(self, *args, **kwargs):
         super(ActivityForm, self).__init__(*args, **kwargs)
 
-        for fieldname in ["sponsors", "rooms", "users_allowed", "groups_allowed"]:
+        for fieldname in ["sponsors", "rooms", "users_allowed", "groups_allowed", "users_blacklisted"]:
             self.fields[fieldname].help_text = None
 
         # Simple way to filter out teachers without hitting LDAP. This
@@ -102,7 +102,9 @@ class ActivityForm(forms.ModelForm):
         # HOWEVER: this will result in LDAP information being queried for *all 1800 users.*
         # We need a better way to accomplish this. The solution below works because it only prints
         # the username field which doesn't require an LDAP query to access.
-        self.fields["users_allowed"].queryset = User.objects.get_students()
+        student_objects = User.objects.get_students()
+        self.fields["users_allowed"].queryset = student_objects
+        self.fields["users_blacklisted"].queryset = student_objects
 
         self.fields["presign"].label = "48 Hour"
         self.fields["default_capacity"].help_text = "Overrides the sum of each room's capacity above, if set."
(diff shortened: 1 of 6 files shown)


intranet/apps/eighth/tests.py
```diff
@@ -70,3 +71,25 @@ def test_signups(self):
...
+
+        act1 = EighthActivity.objects.create(name="Test Activity 1")
+        act1.rooms.add(room1)
+        act1.users_blacklisted.add(user1)
+        schact1 = EighthScheduledActivity.objects.create(activity=act1, block=block1)
+
+        didnt_trigger_error = True
+
+        try:
+            schact1.add_user(user1)
```
pefoley2: Use assertRaises

intranet/apps/eighth/models.py
```diff
@@ -163,6 +163,9 @@ class EighthActivity(AbstractBaseEighthModel):
         groups_allowed
             Individual groups allowed to sign up for this activity. Only takes effect if the activity is
             restricted.
+        users_denied
```
pefoley2: blacklisted

intranet/apps/eighth/models.py
```diff
@@ -928,6 +933,10 @@ def add_user(self, user, request=None, force=False, no_after_deadline=False):
                 acts = EighthActivity.restricted_activities_available_to_user(user)
                 if self.activity.id not in acts:
                     exception.Restricted = True
+            else:
```
pefoley2: Don't have an else here, we want to keep track of all possible errors

intranet/apps/eighth/tests.py
```diff
@@ -70,3 +71,18 @@ def test_signups(self):
...
+        user1 = User.objects.create(username="user1")
+        block1 = EighthBlock.objects.create(date='2015-01-01', block_letter="A")
+        room1 = EighthRoom.objects.create(name="room1")
+
+        act1 = EighthActivity.objects.create(name="Test Activity 1")
+        act1.rooms.add(room1)
+        act1.users_blacklisted.add(user1)
+        schact1 = EighthScheduledActivity.objects.create(activity=act1, block=block1)
+
+        with self.assertRaises(SignupException):
```
pefoley2: Use assertRaisesRegex to make sure the right exception is raised.

intranet/apps/eighth/models.py
ezwang: the else makes it only check if the activity isn't restricted because if the activity is restricted, a blacklist shouldn't be necessary

this shouldn't matter anyways, since if the activity is restricted the field will be disabled so exception.Blacklist will never be set to true

pefoley2: I'm not seeing that blacklisting is disabled for restricted activities. And even if it's unlikely, we might as well support the restricted+blacklisted case :P

jwoglom:
Blacklisting would probably only be used on a activity level, from what I've been told by Mrs Burch. Can you implement it on the EighthActivity level as well, then have the sum of both take effect (unlike the override for capacity, for example)?

jwoglom:
Ignore that, misread how this was implemented.

intranet/apps/eighth/forms/admin/activities.py
```diff
@@ -102,7 +102,9 @@ def __init__(self, *args, **kwargs):
         # HOWEVER: this will result in LDAP information being queried for *all 1800 users.*
         # We need a better way to accomplish this. The solution below works because it only prints
         # the username field which doesn't require an LDAP query to access.
-        self.fields["users_allowed"].queryset = User.objects.get_students()
+        student_objects = User.objects.get_students()
```
pefoley2: Add a FIXME to figure out a cleaner way to do this?

pefoley2:
Minor nit, otherwise LGTM.

intranet/apps/eighth/forms/admin/activities.py
ezwang: Theres a TODO on line 100

pefoley2: Sure, just trying to make it more easily grep-able, could go either way tbh.

PR #401, opened 2016-02-28
fixed table, added date modified for folders

ezwang:
the previous directory and empty directory had only 1 td tag, changed to 3 so it doesn't look weird


pefoley2:
Coverage is irrelevant for this, deferring to @jwoglom on the html.

PR #404, opened 2016-02-29
added client side timeago

ezwang:
if javascript is disabled for some reason, the solution will fall back to the server generated timeago and the title will be the actual time

i would like to add this to the filesystem folders/files, but this would conflict with another one of my pull requests

i didn't include {{ block.super }} in some of the cases since i don't think there was anything to super and when i did it it threw an error


pefoley2:
@ezwang note that you can base this off of another pull request and just rebase it once that one goes in.

PR #407, opened 2016-03-01
added error message on delete permission error

ezwang:
the server would 500 before

### DIFF
--- a/intranet/apps/files/views.py
+++ b/intranet/apps/files/views.py
@@ -401,11 +401,14 @@ def files_delete(request, fstype=None):
         sftp.rmdir(path)
 
     if "confirm" in request.POST:
-        if not is_directory:
-            sftp.remove(filepath)
-        else:
-            rmtree(sftp, filepath)
-        messages.success(request, ("Folder" if is_directory else "File") + " deleted!")
+        try:
+            if not is_directory:
+                sftp.remove(filepath)
+            else:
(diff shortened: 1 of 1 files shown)


intranet/apps/files/views.py
```diff
@@ -401,11 +401,14 @@ def rmtree(sftp, path):
         sftp.rmdir(path)
 
     if "confirm" in request.POST:
-        if not is_directory:
-            sftp.remove(filepath)
-        else:
-            rmtree(sftp, filepath)
-        messages.success(request, ("Folder" if is_directory else "File") + " deleted!")
+        try:
+            if not is_directory:
```
pefoley2: swap this conditional, if is usually better than if not

intranet/apps/files/views.py
```diff
@@ -401,11 +401,14 @@ def rmtree(sftp, path):
...
-            sftp.remove(filepath)
-        else:
-            rmtree(sftp, filepath)
-        messages.success(request, ("Folder" if is_directory else "File") + " deleted!")
+        try:
+            if is_directory:
+                rmtree(sftp, filepath)
+            else:
+                sftp.remove(filepath)
+            messages.success(request, ("Folder" if is_directory else "File") + " deleted!")
```
pefoley2: Don't duplicate the Folder if is_directory conditional

PR #413, opened 2016-03-08
Reversion

pefoley2:
Now merges cleanly with master.
It might make sense to just merge this to start storing history for 8th changes, even if the url itself is undocumented.


jwoglom:
Holdoff until Spring Break to test

pefoley2:
Sure, we'll probably want to add some more UI as well.

pefoley2:
Spring Break Bump.

PR #417, opened 2016-03-19
Styling updates

jacobajit:


pefoley2:
Can you please rebase onto master to remove the unnecessary merge?

jwoglom:
Can you add a screenshot of the navbar change?

jacobajit:
screen shot 2016-03-18 at 9 58 10 pm

pefoley2:
We can ignore the coverage check for this PR, it isn't really relevant.

jwoglom:
Eh, I don't know how to feel about the lack of borders. I think it looks messy without them, to be honest. Without borders all the options run together and you can't differentiate the tappable areas

jacobajit:
I feel that borders should only be used to delineate sections of a menu. This is how most side hamburger menus work on mobile, eg. Google's apps https://www.google.com/design/spec/patterns/navigation-drawer.html#navigation-drawer-specs

jwoglom:
I don't disagree strongly on mobile, but from my quick glance at the code I think it would be applied for desktop as well?

jacobajit:
It would apply on desktop too, but maintaining consistency can't hurt much, and the highlighting hover effect on desktop practically eliminates ambiguity, from my experience. The trend of eliminating dividers seems to exist on desktop too
screen shot 2016-03-18 at 10 15 36 pm
, like with Inbox.

jacobajit:
Also, if we add more items to the nav menu at some point, borders would then help to separate core menu items or categories, while leaving dividers everywhere would serve no purpose (this refers to mobile, a long menu wouldn't work with the current desktop tabs)

ezwang:
As a compromise, I tried making the borders lighter and it doesn't look too bad.

jwoglom:
@ezwang what color are the borders in your screenshot?

ezwang:
the color is rgb(220, 220, 220)

PR #418, opened 2016-03-19
Board

pefoley2:
I commented out the entry in urls.py so that the board module won't actually show up in production.
If this is a feature that we're going to want at some point, then it would be better if it is merged into master now, even if it won't be used for a while rather then having to do another messy and painful rebase down the road.


jwoglom:
Also, on the same page, a swipe from bottom to top to close the block selector, and top to bottom to open it.

jwoglom:
Merge for testing

PR #438, opened 2016-04-05
phone input fixes

ovkulkarni:
2b37b34 None was causing an invalid phone number but it is the default value

### DIFF
--- a/intranet/apps/preferences/fields.py
+++ b/intranet/apps/preferences/fields.py
@@ -12,7 +12,7 @@ class PhoneField(forms.Field):
 
     def __init__(self, *args, **kwargs):
         super(PhoneField, self).__init__(*args, **kwargs)
-        self.validators.append(validators.RegexValidator(r'^\+?1?\d{9,15}$', 'Please enter a valid phone number.'))
+        self.validators.append(validators.RegexValidator(r'^(\+?1?\d{9,15}|\bNone\b)$', 'Please enter a valid phone number.'))
 
     def to_python(self, value):
         "Returns a Unicode object."
--- a/intranet/apps/preferences/fields.py
+++ b/intranet/apps/preferences/fields.py
@@ -10,12 +10,9 @@ class PhoneField(forms.Field):
         'invalid': 'Please enter a valid phone number.'
     }
 
-    def __init__(self, max_length=None, *args, **kwargs):
-        self.max_length = max_length
+    def __init__(self, *args, **kwargs):
         super(PhoneField, self).__init__(*args, **kwargs)
-        self.validators.append(validators.RegexValidator(r'^[0-9]+$', 'Please enter a valid phone number.'))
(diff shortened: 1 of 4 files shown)


intranet/apps/preferences/fields.py
```diff
@@ -25,7 +22,5 @@ def to_python(self, value):
         return value
 
     def widget_attrs(self, widget):
-        attrs = {}
-        if self.max_length is not None:
-            attrs.update({'maxlength': str(self.max_length)})
+        attrs = {"maxlength": "17"}
```
pefoley2: please have a comment to explain this "magic value".

pefoley2:
Don't you need to run makemigrations?

intranet/apps/preferences/fields.py
pefoley2: even better would be a define that is used in all the occurrences.

ovkulkarni: @ezwang

pefoley2:
Test please?

intranet/apps/preferences/fields.py
ezwang: The maximum the regex will accept is 17, I'll add a commit in later.
We should standardize the phone fields and make them all display either "None" or be blank when they are empty. Currently, some of them display "None" when the field is empty and others are blank. I personally think that making them all blank is the best option.

pefoley2: It's possible the None is left over from old db entries?

ovkulkarni: For some reason LDAP is converting the None to a string (which is supposed to happen) as seen [here](https://github.com/tjcsl/ion/blob/master/intranet/apps/preferences/views.py#L97-L102) only for mobile_phone but not any others.

ezwang: I don't think so, I changed the mobile phone field to a phone number, verified it saved, and changed it back to blank, and after a refresh the field displays "None".

ovkulkarni: @ezwang Yes, that's correct, according to the code linked above, it should be set to None.

intranet/apps/preferences/fields.py
```diff
@@ -25,7 +22,5 @@ def to_python(self, value):
         return value
 
     def widget_attrs(self, widget):
```
tjhsst-ion: Codacy Issue found: [Method could be a function](https://www.codacy.com/app/tjcsl/ion/file/2354287857/issues/source?bid=3176160&fileBranchId=3218556#l24)

pefoley2:
Needs a rebase.
Fyi, you can create a temporary branch in the tjcsl/ion repo and then we'll just delete it after this pr is merged.

ovkulkarni:
I'm going to create a new branch as @pefoley2 suggested. Closing PR.

PR #439, opened 2016-04-05
Removed hover/shading on hamburger icon

jacobajit:


jamesthoughton:
@jacobajit I prefer how it is now :(

jacobajit:
IMO it looks more Material and up to date with the revision; @jwoglom thought it was alright.

jamesthoughton:
if you remove it under the hamburger button, then you probably have to remove it from under the username dropdown toggle too (and imo that absolutely should not go)

pefoley2:
If you want to do this, please keep the ui consistent.
What about pulling an inbox and having the color of the hamburger menu slightly change?

jacobajit:
Color change on hover is probably the best solution, for all icons. However, the icons should return to their original color, like the Inbox hamburger. If someone could fix this later

pefoley2:
Sure, but let's keep it as-is until we have a complete fix that keeps the UI internally consistent.
Closing for now, feel free to make a new PR w/ changes to all the UI elements.

PR #496, opened 2016-10-18
dbcache

ovkulkarni:
This allows user information needed for performance to be cached in the database. We should run `./manage.py set_db_cache` once deployed to set the cache, and then per-user cache can be cleared using the GUI. Accessing the cache information should be done by using filter: `User.objects.filter(cache__gender=True)` or `get_or_set_cache`: `User.get_user(username="2019okulkarn").get_or_set_cache('grade_number')`. The first query would get all males, and the second query would get my grade number as an `int`.


pefoley2:
Please fix the unnecessary parens and add a test.

intranet/apps/users/models.py
```diff
@@ -433,7 +436,16 @@ def full_name(self):
         This is required for subclasses of User.
 
         """
-        return self.common_name
+
+        if self.get_from_cache('first_name') is None:
```
pefoley2: Please factor this logic out to a function to avoid duplication.

intranet/apps/users/models.py
```diff
@@ -743,11 +783,41 @@ def birthday(self):
 
     @property
     def is_male(self):
-        return self.sex.lower()[:1] == "m" if self.sex else False
+        if self.get_from_cache('gender') is None:
```
pefoley2: Factor this out as well

intranet/apps/users/models.py
```diff
@@ -1048,8 +1121,11 @@ def is_teacher(self):
             Boolean
 
         """
-
-        return self.user_type == "tjhsstTeacher"
+        if not self.get_from_cache('objectClass'):
```
pefoley2: And this

intranet/apps/users/models.py
```diff
@@ -1554,6 +1644,65 @@ def clear_cache(self):
...
+            "first_name": "first_name",
+            "last_name": "last_name",
+            "graduation_year": "graduation_year"
+        }
+        try:
+            val = getattr(self.cache, attribute)
+            if not val:
+                raise AttributeError
+            return val
+        except AttributeError:
```
pefoley2: don't need the overhead of try-catch, just do something like:
if val: return val
do other stuff

intranet/apps/users/models.py
```diff
@@ -1554,6 +1644,65 @@ def clear_cache(self):
...
+                self.cache = UserCache.objects.create()
+                self.cache.save()
+                self.save()
+            if attribute in mappings:
+                if attribute == "gender":
+                    if self.sex:
+                        self.cache.gender = True if self.sex.lower()[:1] == "m" else False
+                    else:
+                        return None
+                elif attribute == "graduation_year":
```
pefoley2: why not use isdigit instead of hard-coding?

pefoley2:
Also please rebase :)

intranet/apps/users/models.py
ovkulkarni: where should I use isdigit?

intranet/apps/users/models.py
```diff
@@ -1128,18 +1153,17 @@ def is_simple_user(self):
...
     def male(self):
         """Return if the user is male."""
-        return self.sex and self.sex == "M"
+        return self.is_female
 
     @property
     def female(self):
         """Return if the user is female."""
-        return self.sex and self.sex == "F"
+        return self.is_male
```
pefoley2: This needs to be inverted :)

intranet/apps/users/models.py
```diff
@@ -1554,6 +1578,68 @@ def clear_cache(self):
...
             cache.delete(User.create_secure_cache_key(":".join((self.dn, attr))))
+        logger.debug("Clearing cached DB values")
+        if self.cache:
+            UserCache.objects.get(pk=self.cache.pk).delete()
+        self.set_cache()
+
+    def set_cache(self):
+        logger.debug("Setting DB cache using LDAP values")
+        if not self.cache:
+            self.cache = UserCache.objects.create(objectClass=self.user_type, first_name=self.first_name, last_name=self.last_name)
```
pefoley2: This would be more readable if line-wrapped

intranet/apps/users/models.py
```diff
@@ -1554,6 +1578,68 @@ def clear_cache(self):
...
+                        setattr(self.cache, attribute, self.__getattr__(mappings[attribute]))
+                self.cache.save()
+                return self.get_or_set_cache(attribute)
+            elif attribute == "grade_number":
+                logger.debug("Setting grade_number from LDAp")
+                self.cache.grade_number = self.grade.number if self.grade else None
+                self.cache.save()
+                return self.get_or_set_cache(attribute)
+            return None
+        except UserCache.DoesNotExist:
```
pefoley2: The whole thing doesn't need to be in this try-catch block, just have the part that actually throws the exception

intranet/apps/users/tests.py
```diff
@@ -17,3 +22,48 @@ def test_dynamic_groups(self):
...
+        response = self.client.get(reverse('user_profile'), {"clear_cache": 1})
+        self.assertEqual(response.status_code, 302)
+
+    def test_property_without_cache(self):
+        user = self.make_admin()
+        # delete cache object related to user
+        UserCache.objects.filter(user=user).delete()
+        # make sure correct exception is raised when accessing a nonexistent object
+        with self.assertRaises(UserCache.DoesNotExist):
+            user.cache.gender
```
pefoley2: Why not just add a self.assertTrue or something to suppress the warning?

intranet/apps/users/tests.py
```diff
@@ -17,3 +22,48 @@ def test_dynamic_groups(self):
...
+        # make sure user.cache does not exist
+        with self.assertRaises(UserCache.DoesNotExist):
+            user.cache
+        # gender is not accessible, so returns None and sets cache
+        self.assertEqual(user.get_or_set_cache('gender'), None)
+        # make sure cache was set by get_or_set_cache
+        self.assertNotEqual(user.cache, None)
```
pefoley2: assertIsNotNone

intranet/apps/users/tests.py
```diff
@@ -17,3 +22,48 @@ def test_dynamic_groups(self):
...
+        # delete cache object related to user
+        UserCache.objects.filter(user=user).delete()
+        # make sure user.is_male is false and does not fail, i.e checks LDAP and sets cache
+        self.assertEqual(user.is_male, False)
```
pefoley2: assertFalse

intranet/templates/eighth/admin/delinquent_students.html
```diff
@@ -70,7 +71,7 @@
                     <tr>
                         <td>
                             <a href="{% url 'user_profile' delinquent.user.id %}">
-                                {{ delinquent.user.last_name }}, {{ delinquent.user.first_name }}
+                                {{ delinquent.user.last_name' }}, {{ delinquent.user|from_cache:'first_name }}
```
pefoley2: Remove stray '

intranet/templates/eighth/admin/out_of_building_schedules.html
```diff
@@ -1,5 +1,6 @@
 {% extends "eighth/admin/eighth_admin_page_base.html" %}
 {% load static %}
+{% load cache %}
```
pefoley2: Do we actually need this in all files?

intranet/templates/eighth/profile_header.html
```diff
@@ -20,7 +21,7 @@ <h2 title="{{ profile_user.ion_username }} ({{ profile_user.ion_id }})">
                 {% if profile_user.is_student %}
                     {% if profile_user.graduation_year %}
                         {{ profile_user.grade|capfirst }}
-                        (Grade {{ profile_user.grade.number }}: {{ profile_user.graduation_year }})
+                        (Grade {{ profile_user.grade_number' }}: {{ profile_user:from_cache('graduation_year) }})
```
pefoley2: Fix

intranet/apps/templatetags/cache.py
```diff
@@ -0,0 +1,21 @@
+# -*- coding: utf-8 -*-
```
pefoley2: Do we still need this at all?

intranet/apps/users/models.py
```diff
@@ -743,11 +771,11 @@ def birthday(self):
 
     @property
     def is_male(self):
-        return self.sex.lower()[:1] == "m" if self.sex else False
+        return self.get_or_set_cache('gender') is True
 
     @property
     def is_female(self):
-        return self.sex.lower()[:1] == "f" if self.sex else False
+        return self.get_or_set_cache('gender') is False
```
pefoley2: Just do return not

intranet/apps/users/models.py
```diff
@@ -743,11 +771,11 @@ def birthday(self):
 
     @property
     def is_male(self):
-        return self.sex.lower()[:1] == "m" if self.sex else False
+        return self.get_or_set_cache('gender') is True
```
pefoley2: is True is redundant

intranet/apps/users/models.py
pefoley2: What places? The only one i'm seeing that it should be expected is 1610, which could also be refactored to getattr + if, though that's not a big deal

intranet/apps/users/models.py
```diff
@@ -1554,6 +1578,70 @@ def clear_cache(self):
...
+                else:
+                    logger.debug("Setting {} from LDAP".format(attribute))
+                    if str(self.__getattr__(mappings[attribute])).isdigit():
+                        setattr(self.cache, attribute, int(self.__getattr__(mappings[attribute])))
+                    else:
+                        setattr(self.cache, attribute, self.__getattr__(mappings[attribute]))
+                self.cache.save()
+                return self.get_or_set_cache(attribute)
+            elif attribute == "grade_number":
+                logger.debug("Setting grade_number from LDAp")
```
pefoley2: spelling

intranet/templates/dashboard/seniors.html
```diff
@@ -1,3 +1,4 @@
+
```
pefoley2: Please remove unrelated whitespace changes

intranet/apps/users/models.py
ovkulkarni: that would give is_female = True even if they have no gender

intranet/apps/users/models.py
ovkulkarni: see the is_female reply, I'm going to keep it for consistency

intranet/apps/polls/views.py
```diff
@@ -165,7 +167,7 @@ def fmt(num):
 def perc(num, den):
     if den == 0:
         return 0
-    return int(10000 * num / den) / 100
+    return round(num / den * 100, 2)
```
pefoley2: Don't you need to use 100.0 to ensure floating point arith is used?

intranet/apps/polls/views.py
```diff
@@ -247,81 +253,47 @@ def handle_sap(q):
...
+    choices = []
+    for c in q.choice_set.all().order_by("num"):
+        votes = question_votes.filter(choice=c)
+        choices.append(generate_choice(c, votes, total_count, do_gender, show_answers))
+
+    """ Clear vote """
+    votes = question_votes.filter(clear_vote=True)
+    choices.append(generate_choice("Clear vote", votes, total_count, do_gender, show_answers))
+
+    """ Total """
```
pefoley2: Please use ### for this.

PR #502, opened 2016-12-14
remove s from polls

evanyeyeye:


ovkulkarni:
the build is failing because of an error in the dev branch, it should not be fixed in this PR

PR #507, opened 2016-12-29
Change feedback form button to "Submit"

WilliamOConnell:
Changes the label on the button from "Submit Query" to simply "Submit", because from the user's point of view feedback is often not a query.


ezwang:
Looks good to me!

PR #514, opened 2017-03-10
Add workaround for chromium browsers

ebopalisesy:


intranet/apps/signage/views.py
```diff
@@ -113,6 +113,10 @@ def touch_signage(request, sign=None, landscape=False):
     if sign and sign.zoom:
         zoom = sign.zoom
 
+    # Chromium has a weird bug where zoom level of 300% is the same as zoom level 100%
+    if zoom == 3:
+        zoom = '301%'
```
ovkulkarni: how will this work? does signage convert to 300% internally?

ebopalisesy: The CSS tag takes either a float/int or a percent (3.0 or 3 == 300%)

pefoley2:
Why is this an issue? what versions are impacted? is there an upstream bug?

ebopalisesy:
The underlying issue is that Chromium is not handling CSS precedence correctly.  It could be fixed by changing this or by adding an !important in front of the CSS tag in the template.

pefoley2:
Is there an upstream bug?

ebopalisesy:
Not that I could find, no.

pefoley2:
Then file one and put a link to it in the comment explaining why we're doing this.

ebopalisesy:
deleting old workaround and removing the conflicting css tag seems to be another way to fix the issue

PR #515, opened 2017-04-01
Adblock Detect (for April Fool's Day)

cptspck:
On the login page, if users have an adblocker active (prevents ads.js or shows.js from running), then a red banner will display on the top of the saying: 
        We depend on revenue from advertisements to keep ION running for users like you. 
        Without them, we couldn't remain free and accesible to everyone. 
        If you depend on ION, please consider supporting us by disabling your ad blocker.

This only displays on the login page, and wraps behind the login screen on mobile (does not obstruct anything)

### DIFF
--- a/intranet/static/js/ads.js
+++ b/intranet/static/js/ads.js
@@ -1,0 +1,1 @@
+var canRunAds = 1;
--- a/intranet/static/js/showads.js
+++ b/intranet/static/js/showads.js
@@ -1,0 +1,1 @@
+window.canRunAds += 1;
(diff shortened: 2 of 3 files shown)


intranet/static/js/ads.js
```diff
@@ -0,0 +1 @@
+var canRunAds = 1;
```
cptspck: This variable is checked in the login.html file to see of ads.js  and showads.js ran

pefoley2: Shoulda marked it as a global then :)

PR #539, opened 2017-10-26
Miscellaneous Bug Fixes

naitian:


intranet/apps/announcements/notifications.py
```diff
@@ -138,7 +138,7 @@ def announcement_posted_email(request, obj, send_all=False):
                 user_groups = u.groups.all()
                 if any(i in send_groups for i in user_groups):
                     # group intersection exists
-                    em = u.emails[0] if u.emails and len(u.emails) >= 1 else u.tj_email
+                    em = u.emails[0] if u.emails and u.emails.count() >= 1 else u.tj_email
```
ovkulkarni: this still won't work because `u.emails[0]` will error out
it should be `u.emails.first()` or `u.emails.all()[0]`

naitian:
@ovkulkarni fixed

PR #551, opened 2018-02-05
add bus API && vagrant fixes

DylanJones:


Vagrantfile
```diff
@@ -35,7 +35,8 @@ def setup_host
 end
 
 # Make sure the host computer is set up every time a vagrant command is run
-setup_host
+# Comment this back in if you need Kerberos authentication
+# setup_host
```
ezwang: I'm not sure that this should be commented out by default. Having Kerberos authentication is useful when testing features involving LDAP and the only way to login without Kerberos would be to use the master password.

I like what you did with the NFS flag, maybe something similar could be done with Kerberos? You could make a `use_vpn` flag that defaults to true and can be toggled by individual developers.

Vagrantfile
```diff
@@ -66,18 +69,23 @@ Vagrant.configure(VAGRANTFILE_API_VERSION) do |config|
...
     force_group:  'ubuntu',
     perms:        'u=rwX:g=rD:o=rD'
   }
-  config.bindfs.bind_folder "/vagrant-nfs", "/home/ubuntu/intranet",
+  config.bindfs.bind_folder "/vagrant", "/home/ubuntu/intranet",
       force_user: 'ubuntu',
       force_group: 'ubuntu'
-  config.nfs.map_uid = Process.uid
-  config.nfs.map_gid = Process.gid
+  #config.nfs.map_uid = Process.uid
```
ezwang: Should this be dependent on the `use_nfs` flag instead of being commented out?

@naitian might know more about what is necessary for NFS to work correctly.

theo-o:
These changes should really be in separate PRs @DylanJones.

theo-o:
Bus API changes cherry-picked into master/dev as https://github.com/tjcsl/ion/commit/2f08d2a6095382bee722b6774c2e9bfcbb959a60 and https://github.com/tjcsl/ion/commit/2718f402c33d89301fef6820bdb9f192be414f0b. Closing this PR as a result.

PR #552, opened 2018-03-15
Redirect to Eighth Signup during Red lunch

AkashBhave:
Between **12:20** and **12:40** on **Red** days, redirect the user to the eighth period signup page instead of the dashboard.
This will reduce load on the server, because it won't have to load the dashboard **and** the eighth signup page.

Implemented by checking for conditions during login.

### DIFF
--- a/intranet/apps/auth/views.py
+++ b/intranet/apps/auth/views.py
@@ -23,6 +23,7 @@ from ..dashboard.views import dashboard_view, get_fcps_emerg
 from ..schedule.views import schedule_context
 from ..events.models import Event
 from ..users.models import User
+from ..eighth.models import EighthBlock
 
 logger = logging.getLogger(__name__)
 auth_logger = logging.getLogger("intranet_auth")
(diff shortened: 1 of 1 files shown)


jamesthoughton:
I would make this a user preference instead of a mandatory thing

jwoglom:
Just want to point out, irregardless of whether or not this is a good idea to be implemented, that checking whether the name of the schedule entry is equal to "red day" is likely not the correct solution here, given that entries are user-entered and maintained. You'd instead want to check if there was an `EighthBlock` object on that day. Also this as currently implemented would break the eighth period office user, as well as teachers or anyone else who doesn't access the eighth signup page.

AkashBhave:
@jwoglom @jamesthoughton Thanks for the feedback.

Only having this for students is important; I can fix this by adding `request.user.is_student` as another condition in the `if` statement. 

As for checking if it is currently a red day or not, I agree, having a hardcoded value for "Red Day" is not optimal. I looked at the `EighthBlock` class, but couldn't find how to retrieve whether or not it is a red day. Would I use `EighthBlock.objects.get.is_today`?

ezwang:
I would avoid hard coding in the 12:20 to 12:40 times and use the `signup_time` field on the `EighthBlock` object instead. Another possible check could be to check if the current user has signed up for 8th period, and only redirect them if they have not.

ovkulkarni:
+1 to Eric's comment above.

theo-o:
This PR is being replaced by #605. Closing.

PR #556, opened 2018-04-21
Remove dependence on unsupported pip API

theo-o:
In pip>=10.0.0, the pip team got rid of access to pip.download and pip.req  ([https://github.com/pypa/pip/issues/4696](https://github.com/pypa/pip/issues/4696)).  Currently, builds are failing due to this issue.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -1,8 +1,5 @@
 #!/usr/bin/env python
 # -*- coding: utf-8 -*-
-from pip.download import PipSession
-from pip.req import parse_requirements
-
 from setuptools import find_packages, setup
 
 with open('README.rst', 'r') as f:
@@ -10,13 +7,10 @@ with open('README.rst', 'r') as f:
 
 
 def get_requirements():
-    rtd_file = open('docs/rtd-requirements.txt', 'w')
-    for dep in parse_requirements('requirements.txt', session=PipSession()):
-        # FIXME: there should really be a better way to handle this...
-        if dep.req.name not in ['gssapi', 'python-ldap-test']:
-            print(dep.req, file=rtd_file)
-        yield dep.req
-    rtd_file.close()
+    with open('requirements.txt') as req, open('docs/rtd-requirements.txt', 'w') as rtd_file:
+        for dep in req:
+            print(dep.strip(), file=rtd_file)
+            yield dep.strip()
 
 
 setup(


setup.py
```diff
@@ -1,21 +1,21 @@
...
 
+def parse_requirements(filename):
+    lineiter = [line.strip() for line in open(filename)]
+    return [line for line in lineiter if line and not line.startswith("#")]
+
+
 def get_requirements():
     rtd_file = open('docs/rtd-requirements.txt', 'w')
-    for dep in parse_requirements('requirements.txt', session=PipSession()):
-        # FIXME: there should really be a better way to handle this...
```
pefoley2: Glad to see someone finally dealt with my god-awful hack :)

.gitignore
```diff
@@ -30,3 +30,5 @@ fixtures/*.sql
 fixtures/README.txt
 
 *fixtures*.tar*
+.virtualenvs
+.sass-cache
```
pefoley2: These seem unrelated...

intranet/apps/signage/migrations/0015_auto_20180420_1825.py
```diff
@@ -0,0 +1,20 @@
+# -*- coding: utf-8 -*-
```
pefoley2: As do these

setup.py
```diff
@@ -1,21 +1,21 @@
...
-from pip.req import parse_requirements
-
 from setuptools import find_packages, setup
 
 with open('README.rst', 'r') as f:
     long_description = f.read()
 
 
+def parse_requirements(filename):
+    lineiter = [line.strip() for line in open(filename)]
```
pefoley2: Use a with block to avoid leaking the fd?

setup.py
```diff
@@ -1,21 +1,21 @@
...
+    return [line for line in lineiter if line and not line.startswith("#")]
+
+
 def get_requirements():
     rtd_file = open('docs/rtd-requirements.txt', 'w')
```
pefoley2: Why not make this a with block as well to avoid having to call close explicitly?

intranet/apps/signage/migrations/0015_auto_20180420_1825.py
theo-o: The build fails without it.

.gitignore
theo-o: Forgot to drop this.

setup.py
theo-o: Good idea.  Looking at the code again, a compound with statement does make more sense.

intranet/apps/signage/migrations/0015_auto_20180420_1825.py
pefoley2: Sure, but it should really be a separate PR as it's completely unrelated.

theo-o:
Retargeted and rebased

Issue #554, opened 2018-03-30
Upgrade to Django Channels 2


naitian:
This involves some refactoring, because Channels 2 is not backwards-compatible with Channels 1.

https://channels.readthedocs.io/en/latest/one-to-two.html#how-to-upgrade

That said, from what I've read, I don't think we need to change that much.

duvallj:
Files w/ direct dependencies on channels (found using `grep -r channels intranet/*`):
```
intranet/app/bus/consumers.py
intranet/asgi.py
intranet/routing.py
intranet/collected_static/channels/js/websocketbridge.js
```

PR #558, opened 2018-04-23
Add pipenv support

pefoley2:
Because it's patently obvious that nobody is maintaining
requirements.txt, and honestly? I can't really blame them.


Ion.egg-info/requires.txt
```diff
@@ -1,105 +1,95 @@
...
+bcrypt==3.1.4
 beautifulsoup4==4.6.0
-bleach==2.0.0
-certifi==2017.7.27.1
-cffi==1.11.0
-channels==1.1.6
+bleach==2.1.3
+certifi==2018.4.16
+cffi==1.11.5
+channels==2.1.1
```
naitian: This requires some rewriting of the code (as I'm sure is the case with some other major version updates). See #554

pefoley2: Locked the versions of all top-level deps, PTAL.

theo-o:
pipenv is currently not on the Ion team's roadmap as it is currently not a stable and mature packaging solution.

PR #561, opened 2018-05-02
Do not display None as phone number (Fix #535)

### LINKED ISSUE #535: Phone numbers shown confusingly (not huge, but probably not difficult to fix)
The following is shown on all non-freshmen profiles I have checked:

> Phone Number	None (Home Phone) 

The text "None" is also a hyperlink to "tel:None".

Freshmen profiles, however, have the label "Phone Numbers" shown even when there are no phone numbers listed.

theo-o:
A really trivial fix

### DIFF
--- a/intranet/templates/users/profile.html
+++ b/intranet/templates/users/profile.html
@@ -257,11 +257,13 @@
                             </th>
                             <td>
                                 {% for phone in profile_user.phones.all %}
+                                    {% if phone.number %}
                                     <a href="tel:{{ phone.number }}">
                                         {{ phone.number|dashes }}
                                     </a>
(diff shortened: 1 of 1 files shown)


intranet/templates/users/profile.html
```diff
@@ -257,11 +257,13 @@ <h3>
                             </th>
                             <td>
                                 {% for phone in profile_user.phones.all %}
+                                    {% if phone is None %}
```
ovkulkarni: Are you sure this works? the logic seems backwards to me. You should make it `{% if phone %}`.

theo-o:
Note to self: Proofread trivial PRs

PR #562, opened 2018-05-05
Clear unused OAuth tokens

theo-o:
Currently, [https://ion.tjhsst.edu/oauth/authorized_tokens](https://ion.tjhsst.edu/oauth/authorized_tokens) is a huge list of authorized tokens, where almost all of them will never be used again.  Sites using OAuth like Director and TJSTAR don't use tokens after the user logs out.
This PR will use a cron job to delete tokens more than 30 days old.

### DIFF
--- a/cron/oauth-cleartokens.sh
+++ b/cron/oauth-cleartokens.sh
@@ -1,0 +1,7 @@
+#!/bin/bash
+# run at 10pm every day (0 20 * * *)
+
+timestamp=$(date +"%Y-%m-%d-%H%M")
+cd /usr/local/www/intranet3
+./cron/env.sh ./manage.py cleartokens
+echo "Expired OAuth tokens cleared at $timestamp." >> /var/log/ion/oauth-tokens.log
--- a/cron/oauth-cleartokens.sh
+++ b/cron/oauth-cleartokens.sh
@@ -1,5 +1,5 @@
 #!/bin/bash
-# run at 10pm every day (0 20 * * *)
+# run at 10pm every day (0 22 * * *)
 
 timestamp=$(date +"%Y-%m-%d-%H%M")
 cd /usr/local/www/intranet3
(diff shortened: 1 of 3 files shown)


intranet/settings/__init__.py
```diff
@@ -457,7 +457,9 @@ def get_month_seconds():
     'SCOPES': {
         'read': 'Read scope',
         'write': 'Write scope'
-    }
+    },
+    # OAuth refresh tokens expire in 30 days
+    'REFRESH_TOKEN_EXPIRE_SECONDS' : 2592000,
```
naitian: I think it may be a good idea to limit the token expiry to a maximum of a couple of days, since applications should be requesting refresh tokens anyway. That would be more secure and shouldn't affect the end user.

theo-o: There is a different setting for setting the access token expiry [ACCESS_TOKEN_EXPIRE_SECONDS](https://django-oauth-toolkit.readthedocs.io/en/latest/settings.html#access-token-expire-seconds), which is by default set to 10 hrs.  REFRESH_TOKEN_EXPIRE_SECONDS is how long applications can request new access tokens without the user reauthenticating with their password. 
After this PR, the user would need to reauthenticate to Ion with their password every 30 days.  The reason I set it to 30 days is that some applications need access without user authentication for a long period of time (i.e. the unofficial Ion app).

jwoglom: I'd recommend explicitly setting this to ```60*60*24*30``` in the settings file so that it's clear what exactly it represents, rather than an arbitrary number of seconds

cron/oauth-cleartokens.sh
```diff
@@ -0,0 +1,7 @@
+#!/bin/bash
+# run at 10pm every day (0 22 * * *)
+
+timestamp=$(date +"%Y-%m-%d-%H%M")
+cd /usr/local/www/intranet3
```
naitian: `./cron/env.sh` already includes this line.

PR #564, opened 2018-05-25
Hide scrollbars on signage

anonymoose2:
Micro-optimization. No idea if anything else needs to be done.


anonymoose2:
@naitian: Just realized this won't affect iframes, so the announcements page would still have a scrollbar. Would `signage.page.scss` be the place to add it so that doesn't happen?

anonymoose2:
@naitian Is this ready to be merged?

anonymoose2:
Ok, that's weird. Travis fails, but `./setup.py test` locally gives no errors.

theo-o:
Line 251 of `settings/__init__.py` is too long and failing the flake line test (150 char limit).

PR #566, opened 2018-05-25
Add button instead of link

vbhaip:


theo-o:
This does not match our vision for the widget. Closing.

PR #570, opened 2018-07-13
Move  from pycrypto to pycryptodome

theo-o:
pycrypto has had no new releases since October of 2013.  [pycryptodome](https://github.com/Legrandin/pycryptodome) is a fork of pycrypto that is actively maintained. pycryptodome should be a simple drop-in replacement.


pefoley2:
Also relevant: https://nvd.nist.gov/vuln/detail/CVE-2018-6594

PR #580, opened 2018-10-22
Modified files mobile css

DarinMao:
Filesystem cards now take up the full width instead of being 250px

### DIFF
--- a/intranet/static/css/files.scss
+++ b/intranet/static/css/files.scss
@@ -131,6 +131,14 @@ table.directory-list {
     }
 }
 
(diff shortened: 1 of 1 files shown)


intranet/static/css/files.scss
```diff
@@ -131,6 +131,14 @@ table.directory-list {
     }
 }
 
+@media (max-width: 550px) {
```
ovkulkarni:
```suggestion
@media (max-width: 700px) {
```

This is when the search bar drops to a smaller size and approximately when files cards become a single column

PR #586, opened 2019-01-12
Display current signup for each block on eighth activity page

anonymoose2:
This might not be what the page was originally intended to display, but it feels more intuitive from the user's perspective.


intranet/templates/eighth/activity.html
```diff
@@ -125,6 +125,7 @@ <h3>
             <th>Sponsors</th>
             <th>Rooms</th>
             <th></th>
+            <th>Signed up this block?</th>
```
ovkulkarni: Wrap this in `{% if request.user.is_student %}`

intranet/templates/eighth/activity.html
```diff
@@ -125,7 +125,9 @@ <h3>
             <th>Sponsors</th>
             <th>Rooms</th>
             <th></th>
-            <th>Signed up this block?</th>
+            {% if request.user.is_student %}
+                <th>Signed up this block?</th>
+            {% endif}
```
ovkulkarni: `{% endif %}`

PR #587, opened 2019-01-13
In the calendar view, only display 'Passing' period for today

anonymoose2:
Self-explanatory. The calendar view appears to have been added almost as an afterthought, so it's not surprising that there are bugs.

### DIFF
--- a/intranet/static/js/schedule.js
+++ b/intranet/static/js/schedule.js
@@ -78,11 +78,17 @@ $(function() {
         var curDate = formatDate($sch.attr("data-date"));
 
         return blocks.map(function() {
+            var date = curDate;
+            var tbl = $(this).closest("table.bellschedule-table");
+            if(tbl.attr("data-date") != null) {
+                date = formatDate(tbl.attr("data-date"));
+            }
+
             var start = $(this).attr("data-block-start");
-            var startDate = formatTime(start, curDate);
+            var startDate = formatTime(start, date);
 
             var end = $(this).attr("data-block-end");
-            var endDate = formatTime(end, curDate);
+            var endDate = formatTime(end, date);
 
             return {
                 "name": $(this).attr("data-block-name"),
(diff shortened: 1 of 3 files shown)


intranet/static/js/schedule.js
```diff
@@ -100,7 +106,19 @@ $(function() {
     }
 
     getPeriodElem = function(period) {
-        return $(".schedule-block[data-block-order='" + period.order + "']");
+        var elems = $(".schedule-block[data-block-order='" + period.order + "']");
+        if(elems.length > 1) {
+            //We're probably on the calendar view; try and find today's block specifically
+            var dateString = period.end.date.dateFormat("Y-m-d");
+            for(var i = 0; i < elems.length; i++) {
+                var elem = $(elems.get(i));
```
theo-o: Instead of doing these string comparisons, you could just check if the element's grandparent has the today class. https://github.com/tjcsl/ion/blob/1868ef008729dc87fe2e88f9e83ab833b3106331/intranet/templates/schedule/month.html#L11

anonymoose2: @theo-o That would make . I realize not many people leave this page open overnight and expect it to be accurate the next morning, but this method handles that case and it isn't _that_ bad.

anonymoose2: @theo-o That would make the page inaccurate when left open past midnight. I realize not many people leave this page open overnight and expect it to be accurate the next morning (and it isn't, since styles are applied to mark the current day based on the `today` class), but this method handles that case and it isn't _that_ bad.

theo-o: Are we expected to guarantee a user (who leaves the page open overnight) that the schedule will even be correct one for the next day?

anonymoose2: No, but I'm erring on the side of making that the case for the parts I'm changing. The schedule won't be accurate, but it will be partially accurate.

PR #605, opened 2019-06-07
 Redirect to eighth signup page after 12:20 if not signed up

anonymoose2:
This is intended to replace #552 (@AkashBhave has been consulted on this). *Note: (The start time is different -- 12:30 here vs 12:20 there -- but that's minor and can be debated/changed later.)*

History: Last year, someone pointed out the heavy traffic on Ion during Red Day lunches, and I suggested making a "Lite Dashboard" to load on login instead of the main dashboard to help alleviate some of the load. After some discussion, this changed to become redirecting to the eighth signup page, and @AkashBhave heard about it and opened #552.

I've fixed some of the bugs from #552 and added some of the suggestions from the discussion there. I'm not going to cite people here because I don't want to spam them unnecessarily, but if you look at the comments you can probably see where suggestions there match up with code here. If there's a better way to do something that I'm missing, please comment.

As **this** PR is implemented (I don't want to have another discussion about #552), I don't see any problem with automatically redirecting the user with no way to disable it. If somebody logs into Ion at 12:30 and they aren't signed up, then they _should_ be signing up, and strong encouragement to do so is not a bad thing.

Like #552, this checks when the user logs in.


intranet/apps/auth/views.py
```diff
@@ -183,6 +184,18 @@ def post(self, request):
 
             default_next_page = "index"
 
+            if request.user.is_student:
```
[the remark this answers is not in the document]
anonymoose2: Better?

anonymoose2:
I saw #606 and I agree that the user should be able to configure almost all of those. Some of them would be rather annoying if there was no way to turn them off.

However, in this specific case, for the reasons I explained above. I don't really see a problem with automatically redirecting the user. I will, however, implement a config option (default `False`) in this PR in a few minutes.

intranet/apps/auth/views.py
```diff
@@ -183,6 +185,21 @@ def post(self, request):
...
+                if settings.ENABLE_PRE_EIGHTH_REDIRECT:
```
theo-o: Is there a need for an extra layer of indentation? You could just make it `if  request.user is_student and settings.ENABLE_PRE_EIGHTH_REDIRECT:`

intranet/apps/auth/views.py
```diff
@@ -183,6 +185,21 @@ def post(self, request):
...
+                    # Redirect to eighth signup page if the user isn't signed up for eighth period activities
+                    now = timezone.localtime(timezone.now())
+
+                    today_blocks = EighthBlock.objects.filter(date=now.date())
+                    if today_blocks.exists():
+                        for block in today_blocks:
+                            block_start = timezone.make_aware(datetime.combine(now.date(), block.signup_time), now.tzinfo)
+                            time_difference = block_start - now
+                            if timedelta(0) <= time_difference <= timedelta(minutes=10):
```
theo-o:
```suggestion
                            if timedelta(0) <= time_difference <= timedelta(minutes=20):
```
If this happens, we probably want a more significant time period before signups close.

intranet/apps/auth/tests.py
```diff
@@ -27,12 +30,70 @@ def test_login_page(self):
...
+        self.assertTrue(self.does_login_redirect_to(reverse("index")))
+
+    def test_eighth_login_redirect(self):
+        now = timezone.localtime(timezone.now())
+
+        self.login_student()
+        user = User.objects.get(username="awilliam")
+
+        # Don't let blocks created in other tests contaminate these results
+        for block in EighthBlock.objects.all():
```
theo-o: You can make this simpler with `EighthBlock.objects.all().delete()`

intranet/apps/auth/tests.py
```diff
@@ -27,12 +30,70 @@ def test_login_page(self):
         self.assertEqual(self.client.get(reverse("about")).status_code, 200)
         self.assertEqual(self.client.get(reverse("login")).status_code, 200)
 
-    def test_authentication(self):
-        User.objects.get_or_create(username="awilliam")
+    def add_block(self, **args):
+        block = EighthBlock(**args)
```
theo-o: Why not use `objects.create()` for these helper methods?  It saves the objects automatically.

>To create and save an object in a single step, use the create() method.

From: https://docs.djangoproject.com/en/dev/topics/db/queries/#creating-objects

intranet/apps/auth/tests.py
```diff
@@ -27,12 +30,70 @@ def test_login_page(self):
...
+            self.assertTrue(self.does_login_redirect_to(reverse("eighth_signup")))
+            EighthSignup(user=user, scheduled_activity=self.schedule_activity(block_5, activity)).save()
+            self.assertTrue(self.does_login_redirect_to(reverse("index")))
+
+            block_n5 = self.add_block(date=now.date(), signup_time=(now + deltas[-5]).time(), block_letter="C")
+            self.assertTrue(self.does_login_redirect_to(reverse("index")))
+            EighthSignup(user=user, scheduled_activity=self.schedule_activity(block_n5, activity)).save()
+            self.assertTrue(self.does_login_redirect_to(reverse("index")))
+
+            block_15.delete()
```
theo-o: If you want to delete all EighthBlock objects in one swoop, you can use `EighthBlock.objects.all().delete()`

intranet/apps/auth/tests.py
```diff
@@ -27,12 +30,70 @@ def test_login_page(self):
...
+            block.delete()
+
+        deltas = {minutes: datetime.timedelta(minutes=minutes) for minutes in (-5, 5, 15)}
+
+        activity = self.add_activity(name="Test Activity 1")
+
+        with self.settings(ENABLE_PRE_EIGHTH_REDIRECT=True):
+            block_15 = self.add_block(date=now.date(), signup_time=(now + deltas[15]).time(), block_letter="A")
+            self.assertTrue(self.does_login_redirect_to(reverse("index")))
+            EighthSignup(user=user, scheduled_activity=self.schedule_activity(block_15, activity)).save()
```
theo-o: Why not use `EighthSignup.objects.create()`

intranet/apps/auth/tests.py
```diff
@@ -27,12 +30,70 @@ def test_login_page(self):
...
+            self.assertTrue(self.does_login_redirect_to(reverse("index")))
+
+            block_5 = self.add_block(date=now.date(), signup_time=(now + deltas[5]).time(), block_letter="B")
+            self.assertTrue(self.does_login_redirect_to(reverse("eighth_signup")))
+            EighthSignup(user=user, scheduled_activity=self.schedule_activity(block_5, activity)).save()
```
theo-o: Same as above

intranet/apps/auth/tests.py
```diff
@@ -27,12 +30,70 @@ def test_login_page(self):
...
+            self.assertTrue(self.does_login_redirect_to(reverse("index")))
+
+            block_n5 = self.add_block(date=now.date(), signup_time=(now + deltas[-5]).time(), block_letter="C")
+            self.assertTrue(self.does_login_redirect_to(reverse("index")))
+            EighthSignup(user=user, scheduled_activity=self.schedule_activity(block_n5, activity)).save()
```
theo-o: Same as above

PR #625, opened 2019-06-14
Fix bugs accidentally introduced in #622

anonymoose2:
I didn't quite perform date calculations correctly in #622, and the timezone issues are currently causing the build to fail in #587 because (as of the creation of this PR) it's after midnight UTC but before midnight EDT. This PR should fix that.


theo-o:
Also merge to dev not master

PR #641, opened 2019-06-27
Clean up parts of the codebase

anonymoose2:
Part of the #610/#611/#635/#637 chain.

### DIFF
--- a/intranet/apps/announcements/views.py
+++ b/intranet/apps/announcements/views.py
@@ -101,16 +101,14 @@ def request_announcement_view(request):
                 ann = AnnouncementRequest.objects.get(id=obj.id)
                 logger.debug(teacher_objs)
                 approve_self = False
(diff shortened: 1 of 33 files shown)


intranet/apps/announcements/views.py
```diff
@@ -101,6 +101,7 @@ def request_announcement_view(request):
                 ann = AnnouncementRequest.objects.get(id=obj.id)
                 logger.debug(teacher_objs)
                 approve_self = False
+                teacher = None
```
theo-o: The only reason pylint is complaing about it is because https://github.com/tjcsl/ion/blob/0be269ee946a7d21c6444a7753776dbecf681e46/intranet/apps/announcements/views.py#L112
uses teacher which it should not be doing.  We should address that instead of adding this.

PR #643, opened 2019-06-27
Generate a dictionary listing permission names at class definition

anonymoose2:
Currently, the `User.permissions` property lists all the `UserProperties` object's fields and iterates through them to find the permission fields and add them to the dictionary. This should be a more efficient method.


theo-o:
Could we get some tests for this?

PR #649, opened 2019-07-02
Fix the build during July

anonymoose2:
Also took this opportunity to set explicit 00:00:00 times for the `datetime.datetime`s return by `get_date_range_this_year()`.

Note: @theo-o originally tracked down the bug to this function.


theo-o:
Travis appears to be experiencing a backlog of Linux jobs.  I've tested this locally and the Travis build passes. Merging.

PR #658, opened 2019-07-05
Update events app

theo-o:


intranet/apps/events/views.py
```diff
@@ -24,29 +24,34 @@ def events_view(request):
...
         if "approve" in request.POST and is_events_admin:
-            event_id = request.POST.get("approve")
-            event = get_object_or_404(Event, id=event_id)
-            event.rejected = False
-            event.approved = True
-            event.approved_by = request.user
-            event.save()
-            messages.success(request, "Approved event {}".format(event))
+            event_id = get_id(request.POST.get("approve"))
+            if event_id:
```
anonymoose2: Nitpicking here, but I would do:
```suggestion
            if event_id is not None:
```
Since `get_id()` will always either return an int or `None`. I know that ids should never be 0, but this feels a little better.

intranet/apps/events/views.py
```diff
@@ -24,29 +24,34 @@ def events_view(request):
...
         if "reject" in request.POST and is_events_admin:
-            event_id = request.POST.get("reject")
-            event = get_object_or_404(Event, id=event_id)
-            event.approved = False
-            event.rejected = True
-            event.rejected_by = request.user
-            event.save()
-            messages.success(request, "Rejected event {}".format(event))
+            event_id = get_id(request.POST.get("reject"))
+            if event_id:
```
anonymoose2: Same as above.
```suggestion
            if event_id is not None:
```

PR #659, opened 2019-07-16
refactor(users): refactor picture view

theo-o:


intranet/apps/users/views.py
```diff
@@ -94,58 +94,46 @@ def picture_view(request, user_id, year=None):
...
-                image_buffer = io.BytesIO(data)
-            else:
-                image_buffer = io.open(default_image_path, mode="rb")
-
-        response = HttpResponse(content_type="image/jpeg")
-        response["Content-Disposition"] = "filename={}_{}.jpg".format(user_id, year or preferred)
+            data = None
+
+    if data is None:
+        image_buffer = io.BytesIO(data)
```
anonymoose2: I think this line should be switched with line 122.

theo-o: Oh right :man_facepalming:

PR #679, opened 2019-07-20
build: only run pylint on one python version

theo-o:
Since the errors from pylint don't differ (at least substantially) from one Python version to another, we should only run pylint once in the test suite.  This would shave off about 3 minutes from the test matrix.


theo-o:
> LGTM. Do you want to do the same with `flake8` (perhaps make it run on Python 3.7 while Pylint runs on Python 3.6)?

That is smart idea. Will do.

PR #680, opened 2019-07-20
Update Ion logging

theo-o:
We should omit automated user agents in the Ion `app_access.log` file. This PR also sets up weekly rotations of the logs.

### DIFF
--- a/intranet/middleware/access_log.py
+++ b/intranet/middleware/access_log.py
@@ -1,11 +1,13 @@
 import logging
+
 from datetime import datetime
+
+from django.conf import settings
 
 logger = logging.getLogger("intranet_access")
 
 
 class AccessLogMiddleWare:
-
     def __init__(self, get_response):
         self.get_response = get_response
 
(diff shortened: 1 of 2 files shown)


intranet/middleware/access_log.py
```diff
@@ -1,11 +1,13 @@
 import logging
+
```
anonymoose2: Both of these are standard library imports; they don't need a blank line separating them.
```suggestion
```

intranet/middleware/access_log.py
```diff
@@ -27,9 +29,16 @@ def __call__(self, request):
         if isinstance(ip, set):
             ip = ip[0]
 
-        log_line = "{} - {} - [{}] \"{}\" \"{}\"".format(ip, username, datetime.now(), request.get_full_path(), request.META.get(
-            "HTTP_USER_AGENT", ""))
+        user_agent = request.META.get("HTTP_USER_AGENT", "")
+        log_line = '{} - {} - [{}] "{}" "{}"'.format(ip, username, datetime.now(), request.get_full_path(), user_agent)
+
+        loggable = True
+        for user_agent_substring in settings.NONLOGGABLE_USER_AGENT_SUBSTRINGS:
```
anonymoose2: 1. I would wrap this in an `if user_agent:` in case the user agent header is empty.
2. This whole logic just begs to be shortened to `if not any(user_agent_substring in user_agent for user_agent_substring in settings.NONLOGGABLE_USER_AGENT_SUBSTRINGS)`.

PR #682, opened 2019-07-20
Prohibit setting fcpsschools.net as an email

### LINKED ISSUE #599: Prohibit setting fcpsschools.net addresses as email
Includes not sending notifications to fcpsschools.net.  Should also prohibit users from having one as well.

theo-o:
Blocks users from setting an `fcpsschools.net` address via preferences.

Fixes #599

### DIFF
--- a/intranet/apps/preferences/forms.py
+++ b/intranet/apps/preferences/forms.py
@@ -10,18 +10,16 @@ logger = logging.getLogger(__name__)
 
 
 class BusRouteForm(forms.Form):
-
     def __init__(self, *args, **kwargs):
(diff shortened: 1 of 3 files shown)


intranet/apps/preferences/forms.py
```diff
@@ -118,6 +118,13 @@ class Meta:
 
 class EmailForm(forms.ModelForm):
 
+    def clean_address(self):
+        data = self.cleaned_data["address"]
+        if "@fcpsschools.net" in data.lower():
```
anonymoose2:
```suggestion
        if data.lower().strip().endswith("@fcpsschools.net"):
```

PR #684, opened 2019-07-21
Remember if a user has enabled dark mode even after logout

anonymoose2:
This is done by setting a cookie when they log in and checking it when they log out.

### DIFF
--- a/intranet/middleware/dark_mode.py
+++ b/intranet/middleware/dark_mode.py
@@ -1,0 +1,12 @@
+class DarkModeMiddleware:
+    """
+    Set the 'dark-mode-enabled' cookie if the user is logged in and has unlocked dark mode
+    """
+    def __init__(self, get_response):
+        self.get_response = get_response
+
+    def __call__(self, request):
+        response = self.get_response(request)
+        if request.user is not None and request.user.is_authenticated and request.user.dark_mode_properties.dark_mode_unlocked:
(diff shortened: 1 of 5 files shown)


intranet/settings/__init__.py
```diff
@@ -373,6 +373,7 @@
     "corsheaders.middleware.CorsMiddleware",  # CORS headers, for ext. API use
     "simple_history.middleware.HistoryRequestMiddleware",
     "django_prometheus.middleware.PrometheusAfterMiddleware",  # Django Prometheus after
+    "intranet.middleware.dark_mode.DarkModeMiddleware",
```
theo-o: Add a comment

intranet/middleware/dark_mode.py
```diff
@@ -0,0 +1,9 @@
+class DarkModeMiddleware:
```
theo-o: Add a comment on what this middleware does.

intranet/utils/helpers.py
```diff
@@ -162,7 +162,10 @@ def dark_mode_enabled(request):
     if request.GET.get("dark", None):
         return request.GET["dark"] in ["1", "True"]
 
-    if request.user.is_authenticated and request.user.dark_mode_properties.dark_mode_unlocked:
-        return request.user.dark_mode_properties.dark_mode_enabled
+    if request.user.is_authenticated:
+        if request.user.dark_mode_properties.dark_mode_unlocked:
```
theo-o: You could just return dark_mode_unlocked and dark_mode_enabled. This would also allow you to get rid of the `return False` at the end.

PR #689, opened 2019-07-23
Clean up parts of the codebase

anonymoose2:
Part of the #610/#611/#635/#637/#641/#645/#646/#652/#678 chain, but includes a bug fix or two.

### DIFF
--- a/intranet/apps/eighth/models.py
+++ b/intranet/apps/eighth/models.py
@@ -739,11 +739,15 @@ class EighthScheduledActivity(AbstractBaseEighthModel):
         if self.capacity is not None:
             return self.capacity
 
-        if self.activity.default_capacity and not self.rooms.exists():
+        rooms = self.rooms.all()
+        if rooms:
+            return EighthRoom.total_capacity_of_rooms(rooms)
+
+        if self.activity.default_capacity:
             # use activity-level override
             return self.activity.default_capacity
 
-        return EighthRoom.total_capacity_of_rooms(self.get_true_rooms())
+        return EighthRoom.total_capacity_of_rooms(self.activity.rooms.all())
 
     def is_both_blocks(self):
         return self.both_blocks or self.activity.both_blocks
--- a/intranet/apps/eighth/models.py
+++ b/intranet/apps/eighth/models.py
@@ -688,9 +688,8 @@ class EighthScheduledActivity(AbstractBaseEighthModel):
 
     history = HistoricalRecords()
 
-    @property
-    def all_associated_rooms(self):
-        return list(self.rooms.all()) + list(self.activity.rooms.all())
+    def get_all_associated_rooms(self):
+        return (self.rooms.all() | self.activity.rooms.all()).distinct()
 
     @property
     def full_title(self):
--- a/intranet/apps/eighth/models.py
+++ b/intranet/apps/eighth/models.py
@@ -1,6 +1,7 @@
 # pylint: disable=too-many-lines; Allow more than 1000 lines
 import datetime
 import logging
+import string
 
 from django.conf import settings
 from django.contrib.auth import get_user_model
@@ -139,7 +140,7 @@ class EighthRoom(AbstractBaseEighthModel):
     @property
     def formatted_name(self):
         """str: The formatted name of the Room."""
-        if self.name[0].isdigit():
+        if self.name[0] in string.digits:
             # All rooms starting with an integer will be prefixed
             return "Rm. {}".format(self.name)
         if self.name.startswith("Room"):
--- a/intranet/apps/eighth/models.py
+++ b/intranet/apps/eighth/models.py
@@ -804,52 +804,51 @@ class EighthScheduledActivity(AbstractBaseEighthModel):
         """Return whether there are passes that have not been acknowledged."""
         return self.eighthsignup_set.filter(after_deadline=True, pass_accepted=False).exists()
 
-    def get_viewable_members(self, user=None):
-        """Get the list of members that you have permissions to view.
+    def _get_viewable_members(self, user):
+        """Get a (unsorted) QuerySet of the members that you have permission to view.
 
-        Returns: List of members
+        Returns: Unsorted QuerySet of the members that you have permssion to view.
 
         """
-        members = []
-        for member in self.members.all():
-            show = member.can_view_eighth or (user and (user.is_eighth_admin or user.is_teacher or member == user))
+        if user and (user.is_eighth_admin or user.is_eighthoffice or user.is_teacher):
+            return self.members.all()
+        else:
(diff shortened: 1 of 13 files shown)


intranet/apps/eighth/views/admin/attendance.py
```diff
@@ -36,13 +37,13 @@ def delinquent_students_view(request):
     start_date = request.GET.get("start", "")
     end_date = request.GET.get("end", "")
 
-    if not lower_absence_limit.isdigit():
+    if any(c not in string.digits for c in lower_absence_limit):
```
theo-o: Might be worth to factor the digit check into a `util` method since you repeat this so often.

intranet/apps/users/models.py
```diff
@@ -308,6 +309,29 @@ def tj_email(self):
...
+
+        Returns:
+            The first non-TJ email found for a user
+
+        """
+        tj_email = self.tj_email
+        if self.primary_email and self.primary_email != tj_email:
+            return self.primary_email
+
+        for email in self.emails.all():
```
theo-o: Perhaps let the DB handle this by replacing this for loop and the final `return None` with:
```python
return self.emails.exclude(address__iexact=tj_email).first()
```

intranet/apps/eighth/models.py
```diff
@@ -824,28 +826,30 @@ def get_viewable_members_serializer(self, request):
...
-        for member in self.members.all():
-            show = member.can_view_eighth or (user and (user.is_eighth_admin or user.is_teacher or member == user))
-
-            if show:
-                ids.append(member.id)
+        if user and (user.is_eighth_admin or user.is_teacher):
+            return self.members.all()
+        else:
+            ids = []
+            for member in self.members.all():
```
theo-o: You could try to offload a bit of work to the DB:
```python
for member in self.members.exclude(id=user.id).all():
    if member.can_view_eighth:
        ids.append(member.id)
```

intranet/apps/eighth/models.py
```diff
@@ -824,28 +826,30 @@ def get_viewable_members_serializer(self, request):
         Returns: QuerySet
 
         """
-        ids = []
         user = request.user
```
theo-o: Couldn't you factor out the duplicative logic from `get_viewable_members_serializer` and `get_viewable_members`?

intranet/apps/eighth/models.py
anonymoose2: That wouldn't give the required behavior -- if the user is signed up, they should be included too. I'd have to make another query to see if they're signed up and add them separately, which kind of defeats the point.

intranet/apps/eighth/models.py
anonymoose2: I could, but it would be messy and probably slower. Besides, this isn't too much duplication.

intranet/utils/helpers.py
```diff
@@ -1,3 +1,4 @@
+import string
```
theo-o: Alphabetically order

intranet/apps/eighth/models.py
```diff
@@ -802,20 +803,22 @@ def has_open_passes(self):
         """Return whether there are passes that have not been acknowledged."""
         return self.eighthsignup_set.filter(after_deadline=True, pass_accepted=False).exists()
 
+    def _get_viewable_members(self, user):
```
theo-o: Add a docstring

intranet/apps/eighth/models.py
```diff
@@ -802,20 +803,22 @@ def has_open_passes(self):
...
+        else:
+            q = Q(properties__parent_show_eighth=True, properties__self_show_eighth=True)
+            if user:
+                q |= Q(id=user.id)
+            return self.members.filter(q)
+
     def get_viewable_members(self, user=None):
         """Get the list of members that you have permissions to view.
 
         Returns: List of members
```
theo-o:
```suggestion
         Returns: Queryset of members with permission to view
```

intranet/apps/eighth/models.py
```diff
@@ -824,30 +827,22 @@ def get_viewable_members_serializer(self, request):
...
 
         """
-        hidden_members = []
-        for member in self.members.all():
-            show = member.can_view_eighth or (user and (user.is_eighth_admin or user.is_teacher or member == user))
-
-            if not show:
-                hidden_members.append(member)
+        if user and (user.is_eighth_admin or user.is_eighthoffice or user.is_teacher):
+            return self.members.none()
```
theo-o:
```suggestion
            return User.objects.none()
```

intranet/apps/eighth/models.py
```diff
@@ -824,30 +827,22 @@ def get_viewable_members_serializer(self, request):
...
+        else:
+            hidden_members = self.members.all().exclude(properties__parent_show_eighth=True, properties__self_show_eighth=True)
```
theo-o:
```suggestion
            hidden_members = self.members.exclude(properties__parent_show_eighth=True, properties__self_show_eighth=True)
```

intranet/apps/eighth/models.py
```diff
@@ -824,30 +827,22 @@ def get_viewable_members_serializer(self, request):
...
-            if show:
-                ids.append(member.id)
-
-        return get_user_model().objects.filter(id__in=ids)
+        return self._get_viewable_members(request.user)
 
     def get_hidden_members(self, user=None):
         """Get the members that you do not have permission to view.
 
         Returns: List of members hidden based on their permission preferences
```
theo-o: Update docstring

intranet/apps/users/models.py
```diff
@@ -308,6 +309,25 @@ def tj_email(self):
...
+    def non_tj_email(self):
+        """
+        Returns the user's first non-TJ email found.
+
+        If a user has a primary email set and it is not their TJ email,
+        use that. Otherwise, use the first email found that is not their
+        TJ email.
+
+        Returns:
+            The first non-TJ email found for a user
```
theo-o:
```suggestion
            The first non-TJ email found for a user.
```

intranet/apps/eighth/models.py
```diff
@@ -1,4 +1,5 @@
 # pylint: disable=too-many-lines; Allow more than 1000 lines
+import string
```
theo-o: Alphabetical ordering

intranet/apps/eighth/models.py
anonymoose2: `get_viewable_members()` does actually return a `list` -- it calls `sorted()`. Would you recommend using `order_by()` instead?

theo-o: I would recommend returning a list (and therefore using order_by).

PR #691, opened 2019-07-24
Email and profile fixes/improvements

anonymoose2:
The main changes:
* `User.tj_email` and `User.notification_email` now always return string email addresses.
* `User.non_tj_email` now always returns either a string email address or `None`.
* Profile items (websites, phones, email addresses) are now displayed correctly.

Things that were intentionally not changed:
* Certain parts of the codebase, including parts that this PR modifies, hardcode references to `user.tj_email` that perhaps could (should?) be switched to `user.notification_email` were not. That is because the main goal of this PR is fixing bugs. A later PR can easily switch those to `user.notifcation_email`.
* The waitlist functionality (though it is disabled, this is worth mentioning) currently requires `user.primary_email` to be set. While perhaps that could be switched to `user.notification_email`, I do not think that is a good idea. Many people do not check their TJ email address, so they should be required to explicitly set a primary email address to ensure that they will actually get waitlist-related emails.

### DIFF
--- a/intranet/apps/users/models.py
+++ b/intranet/apps/users/models.py
@@ -300,7 +300,7 @@ class User(AbstractBaseUser, PermissionsMixin):
 
         for email in self.emails.all():
             if email.address.endswith(("@fcps.edu", "@tjhsst.edu")):
-                return email
+                return email.address
 
         if self.is_teacher:
             domain = "fcps.edu"
@@ -323,10 +323,12 @@ class User(AbstractBaseUser, PermissionsMixin):
 
         """
         tj_email = self.tj_email
-        if self.primary_email and self.primary_email != tj_email:
-            return self.primary_email
 
-        return self.emails.exclude(address__iexact=tj_email).first()
+        if self.primary_email and self.primary_email.address != tj_email:
+            return self.primary_email.address
+
+        email = self.emails.exclude(address__iexact=tj_email).first()
+        return email.address if email else None
 
     @property
     def notification_email(self):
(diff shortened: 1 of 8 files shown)


intranet/apps/users/models.py
```diff
@@ -300,7 +307,7 @@ def tj_email(self):
 
         for email in self.emails.all():
             if email.address.endswith(("@fcps.edu", "@tjhsst.edu")):
-                return email
+                return email.address
```
theo-o: Maybe return `str(email)`?

intranet/apps/users/models.py
```diff
@@ -323,10 +330,12 @@ def non_tj_email(self):
...
         tj_email = self.tj_email
-        if self.primary_email and self.primary_email != tj_email:
-            return self.primary_email
 
-        return self.emails.exclude(address__iexact=tj_email).first()
+        if self.primary_email_address and self.primary_email_address != tj_email:
+            return self.primary_email_address
+
+        email = self.emails.exclude(address__iexact=tj_email).first()
+        return email.address if email else None
```
theo-o: Maybe return `str(email)` instead of `email.address`?

intranet/apps/users/models.py
```diff
@@ -341,7 +350,11 @@ def notification_email(self):
 
         """
 
-        return self.primary_email or self.emails.first() or self.tj_email
+        if self.primary_email_address:
+            return self.primary_email_address
+
+        email = self.emails.first()
+        return email.address if email and email.address else self.tj_email
```
theo-o: Same

intranet/apps/users/models.py
anonymoose2: I specifically wanted to do `email.address` to be as explicit as possible. In general, I don't like relying on methods like `__str__()`, where the developer can basically pick anything they want it to return and it's hard to argue in a particular direction.

PR #707, opened 2019-07-28
Fixes and Sentry breadcrumbs

anonymoose2:


intranet/apps/schedule/api.py
```diff
@@ -39,6 +39,9 @@ def get_object(self):
             day = Day.objects.get(date=self.kwargs['date'])
             return day
         except Day.DoesNotExist:
-            return None
```
theo-o: A test would be nice for this endpoint.

Issue #710, opened 2019-07-29
Implement notifications for room changes


theo-o:
We should notify users when activities they are signed up or sponsoring have a room change (asynchronously).

anonymoose2:
Requires #604 to be closed first.

Issue #709, opened 2019-07-29
Implement notifications for activity cancellations


theo-o:
We should notify users when an activity they are signed up is cancelled or an activity which they are sponsoring is cancelled (asynchronously).

anonymoose2:
Requires #604 to be closed first.

Issue #156, opened 2015-06-03
Background large operations


jwoglom:
e.x. signing up 2000 students for an activity

alanzhu0:
Why is this closed? There are some more places where this should be done.

PR #713, opened 2019-07-29
Use Celery to send emails

### LINKED ISSUE #604: Use a task queue for sending mass emails
We should use a task queue like celery to perform mass emailing asynchronously.

anonymoose2:
Fixes #604. Makes #156, #709, and #710 possible.

Note that as it is currently set up, this will require `rabbitmq-server` to be installed on both development and production environments, and `scripts/restart_celery.sh` will need to be run 1) before starting the server and 2) any time a file used by Celery is modified (though for development environments, this isn't strictly necessary unless using features that have been offloaded to Celery).

### DIFF
--- a/intranet/celery.py
+++ b/intranet/celery.py
@@ -1,0 +1,9 @@
+import os
+
+from celery import Celery
+
+os.environ.setdefault("DJANGO_SETTINGS_MODULE", "intranet.settings")
+
+app = Celery("intranet")
+app.config_from_object("django.conf:settings", namespace="CELERY")
+app.autodiscover_tasks()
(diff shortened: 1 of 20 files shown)


intranet/celery.py
```diff
@@ -0,0 +1,9 @@
+import os
+
+from celery import Celery
+
+os.environ.setdefault('DJANGO_SETTINGS_MODULE', 'intranet.settings')
```
theo-o: I'm really nitpicking: double quotes instead of single quotes.

anonymoose2: I agree. I was copy/pasting from something else, so it slipped through. Will fix.

PR #735, opened 2019-08-02
Switch from datetime to django.utils.timezone

anonymoose2:
@theo-o: The build passes, and I've gone through and I don't *think* this will break anything, but this is another one that will require careful review before merging.

### DIFF
--- a/intranet/apps/auth/views.py
+++ b/intranet/apps/auth/views.py
@@ -1,6 +1,6 @@
 import logging
 import random
-from datetime import date, datetime, timedelta
+from datetime import date, timedelta
 from dateutil.relativedelta import relativedelta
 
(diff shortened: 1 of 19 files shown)


intranet/apps/auth/views.py
```diff
@@ -1,6 +1,6 @@
 import logging
 import random
-from datetime import date, datetime, timedelta
+from datetime import timedelta
```
theo-o: This is not a blocker to merge, but `django.utils.timezone.timedelta` exists.

anonymoose2: That's only because it's imported from `datetime`: https://docs.djangoproject.com/en/1.11/_modules/django/utils/timezone

Issue #601, opened 2019-05-19
Introduce mandatory announcements


theo-o:
Allow users to request announcements that must be broadcast to the entire school, regardless of preference in settings.  This will likely be restricted to admin, and confirmation via a dialog by an announcements admin should be received before being accepted.

anonymoose2:
The basic functionality for this already exists (`Announcement.notify_email_all`). It was disabled without explanation in 6c9b05e9d94 by commenting out the relevant lines in `intranet/apps/announcements/forms.py`.

PR #748, opened 2019-08-08
Miscellaneous fixes and tweaks

anonymoose2:


intranet/apps/announcements/forms.py
```diff
@@ -29,7 +29,7 @@ def __init__(self, *args, **kwargs):
         )
 
     expiration_date = forms.DateTimeInput()
-    # notify_email_all = forms.BooleanField(required=False, label="Send Email to All")
+    notify_email_all = forms.BooleanField(required=False, label="Send Email to All")
```
theo-o: Does this field actually do anything?

anonymoose2: It sets this field in the model: https://github.com/tjcsl/ion/blob/c0e0b834d923d61811ec5c3763b33d6ae249d7be/intranet/apps/announcements/models.py#L112

Which is checked here: https://github.com/tjcsl/ion/blob/34a2e0c1b4e06433165b4b1b90b32b84c5c5bb0f/intranet/apps/announcements/views.py#L48

And gets passed to this function, which uses it to determine which users to send the email to: https://github.com/tjcsl/ion/blob/34a2e0c1b4e06433165b4b1b90b32b84c5c5bb0f/intranet/apps/announcements/notifications.py#L118

Like [I said in #601](https://github.com/tjcsl/ion/issues/601#issuecomment-517966217), the functionality is all there. It was just disabled.

PR #794, opened 2019-08-29
Fix user already being in group in upload group view

anonymoose2:
## Proposed changes
- Check if the user is in a group before attempting to add them to it

Bug fix.


intranet/apps/eighth/views/admin/groups.py
```diff
@@ -282,8 +284,11 @@ def upload_group_members_view(request, group_id):
             num_users = import_group.user_set.count()
```
theo-o: Your change may change the number of users being added.

PR #806, opened 2019-09-04
Block use of ION in announcements and events

### LINKED ISSUE #805: Consider blocking use of ION in announcements
## Feature description
When writing announcements, submitters should not be allowed to submit announcements (either for approval or directly) that use `ION` (in all caps) to refer to the TJHSST Intranet system.  Instead, we should display a helpful message that blocks them from submitting until they have correct the error:
```
We have detected the use of "ION" in all caps in your announcement. Please correct it to use "Ion". <LINK TO PAGE DESCRIBING WHY IT IS WHAT IT IS>
```

Users are beginning to more actively use "ION" to refer to Ion.  This is incorrect.

Note: We should also consider putting this mandate in our code style guide

-

anonymoose2:
## Proposed changes
- Block the text "ION" in announcements (closes #805)
  * Also clean up some of the announcement templates as necessary to make this change possible
- Block the text "ION" in events (related to #805)
  * Also clean up the edit/modify/request event template as necessary to make this change possible

As per #805:
> Users are beginning to more actively use "ION" to refer to Ion. This is incorrect.

### DIFF
--- a/intranet/static/js/announcement.form.js
+++ b/intranet/static/js/announcement.form.js
@@ -24,6 +24,13 @@ $(function() {
             }
         }
 
+        if($("#id_title").val().match(/\bION\b/) || editor.getData().match(/\bION\b/)) {
+            // People frequently write "ION" instead of the correct spelling, "Ion." See https://github.com/tjcsl/ion/issues/805
+            Messenger().error('We have detected the use of "ION" in all caps in your announcement. Please correct it to use "Ion", <a href="/docs/terminology" style="color:#7F7FFF">the official name</a>.');
+            e.preventDefault();
+            return;
+        }
+
         var button = $("button#submit_announcement");
         button.prop("disabled", true);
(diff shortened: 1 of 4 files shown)


intranet/static/js/announcement.form.js
```diff
@@ -24,6 +24,12 @@ $(function() {
             }
         }
 
+        if($("#id_title").val().match(/\bION\b/) || editor.getData().match(/\bION\b/)) {
```
theo-o: You should probably add a comment explaining what this is doing (for future lookers) and link to the appropriate issue.

intranet/templates/events/add_modify.html
```diff
@@ -30,6 +30,19 @@
                 placeholder: "Links",
                 create: true
             })
+
+            // name of <textarea> is description
+            var editor = CKEDITOR.replace("id_description", {
+                width: "600px"
+            });
+
+            $("form#event_form").bind("submit", function(e) {
```
theo-o: Same

PR #810, opened 2019-09-07
Dark mode tweaks

anonymoose2:
## Proposed changes
-  Change eighth signup block colors to create better contrast between signed up and not signed up blocks
- Use brighter activity icons in dark mode to increase contrast against background

In both cases, this increases contrast in places where there is low contrast.


anonymoose2:
I generated `Eighth-Icons-dark.svg` by copying `Eighth-Icons.svg` and replacing every occurrence of `#484848` with `#B7B7B7` (`48` in hex is 72 in decimal, and 255 - 72 = 183 = `B7` in hex, so `B7` is the rough opposite of `48`).

PR #812, opened 2019-09-09
Improve database query handling

anonymoose2:
## Proposed changes
- Respect user birthday permissions in dashboard birthday widget
- Improve efficiency of eighth lunch redirect check
- Respect `Poll.visible` when checking visible poll list
- Improve efficiency of `User.tj_email` slightly

In addition, the following methods were refactored to reduce the number of queries and had tests added:
- `UserManager.user_with_name()`
- `UserManager.get_teachers()`
- `UserManager.get_teachers_sorted()`
- `User.signed_up_today()`
- `Poll.has_unvoted_polls()`
- `Poll.can_vote()`

Reduces the number of database queries, adds tests, and respects permissions properly.

### DIFF
--- a/intranet/apps/polls/models.py
+++ b/intranet/apps/polls/models.py
@@ -31,7 +31,7 @@ class PollManager(Manager):
 
         """
 
-        return Poll.objects.filter(Q(groups__in=user.groups.all()) | Q(groups__isnull=True))
+        return Poll.objects.filter(visible=True).filter(Q(groups__in=user.groups.all()) | Q(groups__isnull=True))
 
 
 class Poll(models.Model):
--- a/intranet/apps/polls/models.py
+++ b/intranet/apps/polls/models.py
@@ -107,14 +107,10 @@ class Poll(models.Model):
         if not self.in_time_range():
             return False
 
-        if self.groups.count() == 0:
+        if not self.groups.exists():
             return True
 
-        for g in self.groups.all():
-            if g in user.groups.all():
-                return True
-
-        return False
+        return user.groups.intersection(self.groups.all()).exists()
 
     def __str__(self):
         return self.title
--- a/intranet/apps/users/tests.py
+++ b/intranet/apps/users/tests.py
@@ -92,6 +92,43 @@ class UserTest(IonTestCase):
     def test_get_signage_user(self):
         self.assertEqual(get_user_model().get_signage_user().id, 99999)
 
+    def test_get_teachers(self):
+        login_user = self.login()
+        login_user.user_type = "student"
+        login_user.save()
+
+        users = [
+            get_user_model().objects.create(first_name=first_name, last_name=last_name, username=first_name[0].lower() + last_name.lower()[:7],
+                                            user_type="teacher")
+            for first_name, last_name in [
+                ("Michael", "Williams"),
+                ("Miguel", "Wilson"),
+                ("John", "Smith"),
+                ("John", "Adams"),
+                ("Michael", "Adams"),
+                ("Adam", "Smith"),
+                ("Andrew", "Adams"),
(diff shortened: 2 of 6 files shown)


intranet/apps/polls/models.py
```diff
@@ -107,14 +107,10 @@ def can_vote(self, user):
...
-        if self.groups.count() == 0:
+        if not self.groups.exists():
             return True
 
-        for g in self.groups.all():
-            if g in user.groups.all():
-                return True
-
-        return False
+        return user.groups.filter(id__in=self.groups.values_list("id", flat=True))
```
theo-o: You can use `intersect` here (https://docs.djangoproject.com/en/1.11/ref/models/querysets/#intersection)
```suggestion
        return user.groups.intersection(self.groups)
```

intranet/apps/users/tests.py
```diff
@@ -92,6 +93,43 @@ class UserTest(IonTestCase):
...
+            for last_name in [None, "", "Webber"]:
+                if first_name and last_name:
+                    continue
+
+                blank_user.first_name = first_name
+                blank_user.last_name = last_name
+                blank_user.save()
+                self.assertNotIn(blank_user.id, [user.id for user in get_user_model().objects.get_teachers()])
+
+        for user in users:
```
theo-o: Change this to `users.delete()`

intranet/apps/users/tests.py
```diff
@@ -176,6 +214,67 @@ def test_name(self):
...
+        self.assertEqual(get_user_model().objects.user_with_name("Miguel", None).last_name, "Wilson")
+        self.assertEqual(get_user_model().objects.user_with_name("Miguel", None).last_name, "Wilson")
+        self.assertEqual(get_user_model().objects.user_with_name("John", None), None)
+        self.assertEqual(get_user_model().objects.user_with_name("Michael", None), None)
+        self.assertEqual(get_user_model().objects.user_with_name("NOBODY", None), None)
+        self.assertEqual(get_user_model().objects.user_with_name(None, "Adams"), None)
+        self.assertEqual(get_user_model().objects.user_with_name(None, "Smith"), None)
+        self.assertEqual(get_user_model().objects.user_with_name(None, "Williams").first_name, "Michael")
+
+        for user in users:
```
theo-o: `users.delete()`

intranet/apps/users/tests.py
```diff
@@ -196,6 +295,94 @@ def test_notification_email(self):
...
+        poll.delete()
+
+    def test_signed_up_today(self):
+        user = self.login()
+        user.user_type = "student"
+        user.save()
+
+        EighthBlock.objects.all().delete()
+
+        assert user.signed_up_today()
```
theo-o: Is there a reason why you use plain `assert`s instead of `self.assertTrue()` and `self.assertFalse()`?

intranet/apps/users/tests.py
anonymoose2: I normally would, but here `users` is a list.

intranet/apps/users/tests.py
anonymoose2: Same here.

intranet/apps/users/tests.py
theo-o: Ignore me :man_facepalming:

intranet/apps/users/tests.py
anonymoose2: Force of habit. I've recently been writing tests for `pytest`, which allows you to use plain `assert`s. (You still get verbose messages because it looks through the stack frame and prints out helpful information.)

If you ever notice that I've done this again, then it's unintentional and you should make me fix it.

PR #820, opened 2019-09-20
Add special message for users of the unofficial iOS app

anonymoose2:
## Proposed changes
- Add special message for users of the unofficial iOS app

People are complaining about the unofficial iOS app not working. We should head off these complaints by adding a specific message explaining that they should not use it.

### DIFF
--- a/intranet/templates/oauth2_provider/authorize.html
+++ b/intranet/templates/oauth2_provider/authorize.html
@@ -61,6 +61,9 @@
                 </div>
             </form>
 
+        {% elif request.GET.client_id in DJANGO_SETTINGS.IOS_APP_CLIENT_IDS %}
+            <h2 style="height: initial; white-space: initial; padding: 5px 10px;">Please do not use the unofficial Ion iOS app</h2>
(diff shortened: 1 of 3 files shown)


intranet/templates/oauth2_provider/authorize.html
```diff
@@ -61,6 +61,9 @@ <h3 class="block-center-heading">A third-party application wants access to your
                 </div>
             </form>
 
+        {% elif request.GET.client_id in DJANGO_SETTINGS.IOS_APP_CLIENT_IDS %}
```
theo-o: How about if `client_id`does not exist in`request.GET`?

anonymoose2: Testing indicates that it fails gracefully. The Django template language is designed to fail silently in a lot of cases like this (which sometimes is helpful and sometimes is not :slightly_smiling_face:).

More specifically, it appears to evaluate to `None` (adding `None to `IOS_APP_CLIENT_IDS` and visiting the page without passing `client_id` causes the message to appear).

PR #837, opened 2019-10-08
Miscellaneous fixes

anonymoose2:
## Proposed changes
- Attempt to prevent race conditions in eighth admin forms
- In dark mode, show every other eighth activity in a lighter color to match light mode

Both small tweaks.

### DIFF
--- a/intranet/apps/eighth/views/admin/groups.py
+++ b/intranet/apps/eighth/views/admin/groups.py
@@ -637,7 +637,7 @@ def add_member_to_group_view(request, group_id):
 
     if "user_id" in request.POST:
         user_ids = request.POST.getlist("user_id")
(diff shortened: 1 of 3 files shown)


intranet/apps/eighth/views/admin/groups.py
```diff
@@ -640,8 +640,9 @@ def add_member_to_group_view(request, group_id):
         user_objects = get_user_model().objects.filter(id__in=user_ids)
         next_url += "?"
         for user in user_objects:
-            user.groups.add(group)
-            user.save()
+            if not user.groups.filter(id=group.id).exists():
```
theo-o: Couldn’t you exclude users on line 642, instead of making new DB queries.

PR #838, opened 2019-10-08
Update printing

theo-o:
## Proposed changes
- Starts adding docstrings
- Update some printing messages
- Modify restrictions on printing
- Update "supported" file formats

Improves printing code.

### DIFF
--- a/intranet/templates/printing/print_form.html
+++ b/intranet/templates/printing/print_form.html
@@ -9,22 +9,22 @@
     </tr>
     </table>
     <br>
-    <p><span style="font-style:italic">Note</span>: Documents of type .doc and .docx are known to often contain errors when printed, so, if possible, print PDF documents for accurate results.</p>
+    <p><span style="font-style:italic">Note</span>: Due to the proprietary nature of the .doc and .docx file format, those formats are unsupported. Print PDF documents for accurate results.</p>
     <b>The following file formats are supported:</b>
     <ul>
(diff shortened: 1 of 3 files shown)


intranet/templates/printing/print_form.html
```diff
@@ -9,22 +9,22 @@
     </tr>
     </table>
     <br>
-    <p><span style="font-style:italic">Note</span>: Documents of type .doc and .docx are known to often contain errors when printed, so, if possible, print PDF documents for accurate results.</p>
+    <p><span style="font-style:italic">Note</span>: Due to the proprietary nature of the .doc and .docx file format, those formats are unsupported. Print PDF documents for accurate results.</p>
```
anonymoose2: I'm not sure that marking the proprietary Word formats as *unsupported* is a good idea. I think we should definitely have a warning about errors, but I think we should continue to list them in the supported list. Since we fixed the ZIP identification issues, I don't think we've had any complaints, and even the inevitable formatting issues aren't a huge problem for most people.

Also, if we *do* mark Word files as unsupported, we should probably explain the formatting issues instead of saying they're proprietary, or people will contest this.

theo-o: Reworded in https://github.com/tjcsl/ion/pull/838/commits/63482c6569811d5647f11b1d80e21db26d1d9ed1

PR #843, opened 2019-10-11
Remove privacy form

theo-o:
## Proposed changes
- Disables privacy options form.
- 
- 

The form is currently obsolete.

### DIFF
--- a/intranet/apps/preferences/views.py
+++ b/intranet/apps/preferences/views.py
@@ -80,13 +80,11 @@ def get_preferred_pic(user):
 
 def save_preferred_pic(request, user):
     preferred_pic = get_preferred_pic(user)
-    logger.debug(preferred_pic)
     preferred_pic_form = PreferredPictureForm(user, data=request.POST, initial=preferred_pic)
     if preferred_pic_form.is_valid():
         logger.debug("Preferred pic form: valid")
         if preferred_pic_form.has_changed():
             fields = preferred_pic_form.cleaned_data
-            logger.debug(fields)
             if "preferred_photo" in fields:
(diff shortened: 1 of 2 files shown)


intranet/apps/preferences/views.py
```diff
@@ -337,9 +337,11 @@ def preferences_view(request):
             preferred_pic_form = PreferredPictureForm(user, initial=preferred_pic)
             bus_route_form = BusRouteForm(initial=bus_route)
 
+            """
             privacy_options = get_privacy_options(user)
-            logger.debug(privacy_options)
             privacy_options_form = PrivacyOptionsForm(user, initial=privacy_options)
+            """
+            privacy_options_form = None
```
anonymoose2: I would disable the call to `save_privacy_options()` above, if only to be consistent. We might also want to add comments explaining why this is disabled.

intranet/templates/preferences/preferences.html
```diff
@@ -205,17 +205,18 @@ <h3>Personal Information</h3>
...
-            <p>Since the eighth period office and TJ faculty can always see your pictures, it is recommended that you choose your preferred picture even if you disable "Show Pictures" in Privacy Options.</p>
+            <p>Since the eighth period office and TJ faculty can always see your pictures, it is recommended that you choose your preferred picture.</p>
 
             {% for choice in preferred_pic_form.preferred_photo %}
             {{ choice.tag }} {{ choice.choice_label }}<br>
             {% endfor %}
             <br>
             {% endif %}
 
+            {% comment %}
```
anonymoose2: Again, we might want to add comments explaining why this is disabled.

PR #849, opened 2019-10-12
Fix issues with session management page

### LINKED ISSUE #846: Change reauthentication implementation to avoid security issues when session management page is added
## Topic of discussion
Currently, when the user is required to re-authenticate to Ion to perform certain maintenance tasks, the `reauthenticated` session key is set to `True`. This then enables users to view pages that require reauthentication until their session expires.

The problem is that the session management page proposed in #559/#771 and implemented in #839 would reduce the security of this reauthentication method, since users would be able to perform these maintenance tasks on any of their trusted devices for several days after having reauthenticated only once. We should fix this so users have to re-authenticate every few hours, at least.

## Proposed options/solutions
*First, whatever the re-authenticate expiration timeout is and however it is implemented, it should be defined in `settings/__init__.py`. so it can be changed easily.*

1. Instead of setting the `reauthenticated` key, set a `reauthenticated_at` (or similar) key to the current time (probably just `time.time()` for simplicity). The `reauthentication_required` decorator can then check if the amount of time that has elapsed since this was set is less than the expiration timeout.
2. Set a separate cookie with the current time. The problem here is that we would need to do a number of things in order to properly secure it -- like signing the cookie so we know it's from a trusted source -- and at that point we're almost reimplementing Django sessions.

We also need to decide on whether or not this should auto-renew every time the user visits Ion once they have re-authenticated.

## Personal opinion (if applicable)
I am in favor of simply setting a different field in the session. I also think that auto-renewing this every time the user visits Ion would introduce additional security issues, so we should just set it once when they re-authenticate.

anonymoose2:
## Proposed changes
- Change reauthentication implementation so it doesn't conflict with the new session management page (closes #846)
- Delete and block duplicate `TrustedSession` objects
- Don't attempt to create duplicate `TrustedSession`s

Fixes a conflict between the reauthentication and session management systems and prevents duplicate `TrustedSession` objects from being created.


intranet/apps/auth/decorators.py
```diff
@@ -51,7 +53,10 @@ def inner(*args, **kwargs):
 def reauthentication_required(wrapped):
     def inner(*args, **kwargs):
         request = args[0]  # request is the first argument in a view
-        if request.session.get("reauthenticated", False):
+        if (
+            "reauthenticated_at" in request.session
+            and 0 <= (time.time() - request.session["reauthenticated_at"]) <= settings.REAUTHENTICATION_EXPIRE_TIMEOUT
```
theo-o: There is no validation that `request.session["reauthenticated_at"]` is an integer.

theo-o: Disregard this

anonymoose2: I already fixed it.

PR #853, opened 2019-10-13
Log cleanup

theo-o:
## Proposed changes
- Cleans up unneeded debug log statements
- Refactors an if statement

We have a lot of log statements filling up our prod logs that are just taking up space. Log statements should provide context to what they are logging instead of just dumping information.

### DIFF
--- a/intranet/apps/emerg/views.py
+++ b/intranet/apps/emerg/views.py
@@ -67,6 +67,7 @@ def check_emerg():
 def get_emerg_result():
     """Run the fetch command from FCPS."""
     status, message = check_emerg()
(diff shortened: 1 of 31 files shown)


intranet/apps/emerg/views.py
```diff
@@ -66,7 +66,6 @@ def check_emerg():
 
 def get_emerg_result():
     """Run the fetch command from FCPS."""
-    logger.debug("Fetching emergency info from FCPS")
```
anonymoose2: I would actually keep this. It might be helpful to know when we last fetched emergency info from FCPS.

PR #868, opened 2019-10-30
Add "features" app for information on new features

### LINKED ISSUE #860: Add method for conveying information about new features to users in a non-annoying way
## Feature description
Add a generalized mechanism for conveying information about new features to users that allows for recording of whether each user has seen the feature in order to prevent showing a user the same feature over and over again.

Currently, the only mechanism for recording whether users have seen the announcements about new features to users is a hacky mechanism that sets a session variable. In addition, announcements are usually hard-coded and removed manually when no longer applicable, which is not good practice.

- A new `features` app that records a list of new feature announcements, as well as whether each user has seen a particular announcement.
  - Required features (no pun intended) of this app:
    - Feature announcements should have an "activation date" before which they should not be displayed.
    - Feature announcements should have an "expiration date" after which they should not be displayed.
    - It should be possible somehow to choose the "context" in which this feature is displayed (this will most likely only ever be `login`, `index`, or `eighth_signup`, so perhaps those can be hard-coded).
    - Models in the `features` app should be cached aggressively, and caches should be properly invalidated when necessary.
    - The `features` app should be properly tested to avoid contributing to #361.
    - (minor) As per #814, all URLs in the `features` app should be namespaced properly.
  - Note that this implementation would still enable features to be added without manual intervention in production through the use of data migrations to create objects.

anonymoose2:
## Proposed changes
- Add "features" app for information on new features (closes #860)
- Remove the temporary dark mode feature announcement on the dashboard

See #860


intranet/apps/features/views.py
```diff
@@ -0,0 +1,22 @@
...
+
+# Create your views here.
+
+
+@require_POST
+@login_required
+@deny_restricted
+def dismiss_feat_announcement_view(request, feat_announcement_id):
+    """Dismiss the specified FeatureAnnouncement for the current user."""
+    feat_announcement = FeatureAnnouncement.objects.get(id=feat_announcement_id)
```
theo-o: We should handle an invalid feat_announcement_id.

theo-o:
@anonymoose2, you should rebase against the latest dev to incorporate changes merged with https://github.com/tjcsl/ion/pull/877.

PR #871, opened 2019-10-31
Darken borders in dark mode

anonymoose2:
## Proposed changes
- Darken borders in dark mode

A lot of people seem to be requesting this.

### DIFF
--- a/intranet/static/css/_colors.scss
+++ b/intranet/static/css/_colors.scss
@@ -1,1 +1,2 @@
 $grey: rgb(72, 72, 72);
+$darkborder: #888;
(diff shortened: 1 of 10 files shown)


theo-o:
I would just like to register my opinion that I don't really see the point in this new behavior.

anonymoose2:
People don't seem to like the bright borders (they aren't quite white, but they're pretty close). I was against this change for a while, but having seen the two side by side I have to admit that the borders do make Ion look a crowded and distract from the content. I think we should definitely keep the borders, but they _are_ a little too bright.

Issue #772, opened 2019-08-15
Add a logout button next to nav dropdown


theo-o:
## Feature description
Adding a logout button (either the power button icon or arrow pointing out icon would work) to the left of the navbar that when clicked logs a user out.

## Rationale for adding feature
Currently, it takes two clicks to logout from Ion, and we should be making it easy for students to be secure and protect their Ion sessions.  Eliminating an extra click would do so and would not be a major disruption to our nav UI.

## Suggested implementation(s) (if applicable)
-

jatinAroraGit:
Hey there , I can work on this issue.

theo-o:
> Hey there , I can work on this issue.

Hi @jatinAroraGit, thank you for your interest in contributing to Ion.  Ion is an internal application that we have made open source. Due to program constraints, we are unable to accept pull requests from persons not part of the TJ community.

PR #874, opened 2019-11-01
feat(nav): add logout button

Laur04:
## Proposed changes
- 
- 
-


anonymoose2:
1. Please mark this as closing #772 (i.e. a second line in the commit message that says `Closes #772`).
2. It's invalid HTML to put anything other than a `<li>` in a `<ul>`. In addition, to get the alignment and everything right, you'll need to make some substantial changes.
   Specifically, you'll probably want to remove the `right` class from the `<ul>` and create a `<div>` around it with the `right` class instead. Then you can `float` the `<ul>` to thr left and put the logout button after it. You'll also probably need to adjust padding more.
3. Please keep the existing logout button on the dropdown button. People expect it to be there by now.
4. We don't need the text "Logout" on the new login button.

PR #875, opened 2019-11-01
feat(nav): add logout button

### LINKED ISSUE #772: Add a logout button next to nav dropdown

Laur04:
## Proposed changes
- add logout button
- 
- 

Closes #772

### DIFF
--- a/intranet/apps/features/models.py
+++ b/intranet/apps/features/models.py
@@ -78,6 +78,7 @@ class FeatureAnnouncement(models.Model):
             (clicked the close button). Used to hide feature announcements the user has dismissed.
 
     """
+
     objects = FeatureAnnouncementQuerySet.as_manager()
 
     # Both INCLUSIVE


intranet/static/css/page_base.scss
```diff
@@ -43,6 +43,7 @@ h1 {
     .right {
         -webkit-user-select: none;
         user-select: none;
+        padding-right:20px;
```
anonymoose2: I would prefer to have this as an inline style on the `<div>`, since we may have other uses of the `right` class that do not need padding.

PR #877, opened 2019-11-02
Enforce static/template file formatting (currently just no trailing whitespace)

anonymoose2:
## Proposed changes
- Add script to format static/template files (currently just removes trailing whitespace)
- Remove trailing whitespace from static/template files
- Add static/template file format script to build

We enforce strict formatting for code, but not for static files and templates. We should begin moving in that direction.

### DIFF
--- a/scripts/static_templates_format.sh
+++ b/scripts/static_templates_format.sh
@@ -1,0 +1,7 @@
+#!/bin/bash
+cd "$(dirname -- "$(dirname -- "$(readlink -f -- "$0")")")"
+
+# Remove trailing whitespace for all JS/CSS/SCSS/txt/JSON/SVG static files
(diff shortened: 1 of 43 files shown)


scripts/static_templates_format.sh
```diff
@@ -0,0 +1,5 @@
+#!/bin/bash
+cd "$(dirname -- "$(dirname -- "$(readlink -f -- "$0")")")"
+
+find intranet/static -name vendor -prune -o -type f '(' -name '*.js' -o -name '*.css' -o -name '*.scss' -o -name '*.txt' -o -name '*.json' -o -name '*.svg' ')' -exec sed -i 's/\s\+$//' '{}' ';'
```
theo-o: We should have a comment describing what this command is intending to do.

scripts/static_templates_format.sh
```diff
@@ -0,0 +1,5 @@
...
+find intranet/templates -type f -exec sed -i 's/\s\+$//' '{}' ';'
```
theo-o: Same here.

PR #883, opened 2019-11-09
refactor: allow configuration of Eighth Period Coordinator name

theo-o:
## Proposed changes
- Allow configuration of 8th period coordinator's name
- 
- 

Currently, the name is hardcoded which is suboptimal.

### DIFF
--- a/intranet/apps/eighth/views/attendance.py
+++ b/intranet/apps/eighth/views/attendance.py
@@ -1,5 +1,6 @@
 import csv
 import logging
+from html import escape
 from io import BytesIO
 
 from cacheops import invalidate_obj
@@ -12,6 +13,7 @@ from reportlab.lib.units import inch
 from reportlab.platypus import PageBreak, Paragraph, SimpleDocTemplate, Spacer, Table, TableStyle
 
 from django import http
+from django.conf import settings
 from django.contrib import messages
 from django.contrib.auth import get_user_model
 from django.contrib.auth.decorators import login_required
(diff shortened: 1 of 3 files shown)


intranet/apps/welcome/views.py
```diff
@@ -16,7 +18,10 @@ def student_welcome_view(request):
 
 @login_required
 def teacher_welcome_view(request):
-    context = {"first_login": request.session["first_login"] if "first_login" in request.session else False}
+    context = {
+        "first_login": request.session["first_login"] if "first_login" in request.session else False,
+        "coordinator": settings.EIGHTH_COORDINATOR_NAME,
```
anonymoose2: Since d277b79a8add567026888cd10ca20219965d921b, we export the `settings` module as the context variable `DJANGO_SETTINGS`. Perhaps we should use that instead of adding it to the context here?

intranet/apps/eighth/views/attendance.py
```diff
@@ -598,12 +599,15 @@ def generate_roster_pdf(sched_act_ids):
...
+        # NOTE: We should really not be writing raw HTML
         instructions = """
         <b>Highlight or circle</b> the names of students who are <b>absent</b>, and put an <b>"X"</b> next to those <b>present</b>.<br/>
         If a student arrives and their name is not on the roster, please send them to the <b>8th Period Office</b>.<br/>
         If a student leaves your activity early, please make a note. <b>Do not make any additions to the roster.</b><br/>
-        Before leaving for the day, return the roster and any passes to 8th Period coordinator, Catherine Massar's mailbox in the
-        <b>main office</b>. For questions, please call extension 5046 or 5078. Thank you!<br/>"""
+        Before leaving for the day, return the roster and any passes to 8th Period coordinator, {}'s mailbox in the
+        <b>main office</b>. For questions, please call extension 5046 or 5078. Thank you!<br/>""".format(
+            settings.EIGHTH_COORDINATOR_NAME
```
anonymoose2: Perhaps we should escape this string using one of the following:
- [`django.utils.html.format_html()`](https://docs.djangoproject.com/en/2.2/ref/utils/#django.utils.html.format_html) 
- [`html.escape()`](https://docs.python.org/3/library/html.html#html.escape)

theo-o: This should probably be incorporated in an overhaul of this report mess.

anonymoose2: Still, it might be a good idea to add `html.escape()` as a temporary solution.

Issue #804, opened 2019-09-02
Await django-prometheus that supports Django 2.2


theo-o:
There is no marking of what Django versions are officially supported although the build runs with Django 2.0 since https://github.com/korfuri/django-prometheus/commit/d0d171ae5c2cddf230ea6f40599d9f47bb511754.

According to `https://github.com/korfuri/django-prometheus/pull/107`, there appear to be failures with the manual tests with Django 2.2.

theo-o:
It appears that django-prometheus (as released on September 28th) supports Django 2.2 (see [setup.py](https://github.com/korfuri/django-prometheus/tree/fe0902f0f78933f020d6e2c039a40c28814414a7) and [.travis.yml](https://github.com/korfuri/django-prometheus/blob/fe0902f0f78933f020d6e2c039a40c28814414a7/.travis.yml#L21))

PR #885, opened 2019-11-09
Package upgrades

theo-o:
## Proposed changes
- Update a lot of packages
- 
- 

Many of our direct dependencies are out-of-date.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -12,7 +12,7 @@ contextlib2==0.6.0
 cryptography==2.8
 decorator==4.3.2
 Django==1.11.26
-django-cacheops==4.1
+django-cacheops==4.2
 django-cors-headers==3.0.2
 django-debug-toolbar==2.0
 django-extensions==2.1.9
--- a/requirements.txt
+++ b/requirements.txt
@@ -9,7 +9,7 @@ certifi==2019.09.11
 channels==2.3.1
 channels-redis==2.4.1
 contextlib2==0.6.0
-cryptography==2.7
+cryptography==2.8
 decorator==4.3.0
 Django==1.11.23
 django-cacheops==4.1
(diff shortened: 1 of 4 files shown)


requirements.txt
```diff
@@ -1,56 +1,56 @@
...
-django-extensions==2.1.9
+django-extensions==2.2.5
 django-formtools==2.1
 django-inline-svg==0.1.1
 django-maintenance-mode==0.13.3
-django-oauth-toolkit==1.1.2
+django-oauth-toolkit==1.1.3
 django-pipeline==1.6.14
-django-prometheus==1.0.15
+django-prometheus==1.1.0
```
anonymoose2: The commit adding this change should be marked as fixing #804.

requirements.txt
```diff
@@ -1,56 +1,56 @@
...
+python-dateutil==2.8.1
 python-gssapi==0.6.4
 python-magic==0.4.15
-reportlab==3.5.23
+reportlab==3.5.26
 requests==2.22.0
-requests-oauthlib==1.2.0
-sentry-sdk==0.11.2
+requests-oauthlib==1.3.0
+sentry-sdk==0.13.2
```
anonymoose2: As of [version 0.12.0](https://github.com/getsentry/sentry-python/releases/tag/0.12.0), Sentry requires `aiocontextvars` when using Channels and running on Python 3.5 or 3.6. (The release notes also mention that previously it leaked memory in these scenarios, so this may be the cause of some of the memory leaks noticed in production earlier today).

`sentry-sdk` does not automatically pull this in (my guess is they don't want to pull it in for projects that don't need it), so we will have to install it manually on versions < 3.7. I think this will work (untested):
`aiocontextvars==0.2.2; python_version < "3.7"`

theo-o: Ah I missed that when I was skimming through change logs.

PR #897, opened 2019-11-16
Improve login theme system

anonymoose2:
## Proposed changes
- Improve login theme system

It is not easy to tell which login theme is active if changes need to be made other than including the CSS/JS files (such as with halloween mode, which needs dark mode to be enabled). This fixes that issue.

### DIFF
--- a/intranet/apps/auth/helpers.py
+++ b/intranet/apps/auth/helpers.py
@@ -1,9 +1,11 @@
 import logging
 import re
+from typing import Dict
 
 import pexpect
 
 from django.conf import settings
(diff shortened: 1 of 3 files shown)


intranet/apps/auth/helpers.py
```diff
@@ -45,3 +46,28 @@ def change_password(form_data):
     if exitstatus == 0:
         return {"unable_to_set": False}
     return {"unable_to_set": True}
+
+
+def get_login_theme_name():
```
theo-o: It would be nice to have a docstring

PR #899, opened 2019-11-16
Add signage eighth period page

### LINKED ISSUE #822: Resurrect 8th period page on signage
## Feature description
On Signage, we should have a tab that displays the eighth period activities coming up. This just needs to contain name, sponsor, and room. This does not need to contain info like capacity. We may be able to draw inspiration from the original signage implementation.  We also do not necessarily need the ability to switch days.

If possible, we should cache this page.

Request by eighth period office.

-

anonymoose2:
## Proposed changes
- Add signage eighth period page (closes #822)

Requested by eighth period office (see #822)

### DIFF
--- a/intranet/apps/signage/views.py
+++ b/intranet/apps/signage/views.py
@@ -1,9 +1,14 @@
 import logging
 
+from django import http
 from django.conf import settings
 from django.shortcuts import get_object_or_404, render
 
+from ...utils.serialization import safe_json
+from ..eighth.models import EighthBlock
+from ..eighth.serializers import EighthBlockDetailSerializer
 from ..schedule.views import schedule_context
+from ..users.models import User
 from .models import Sign
 
 logger = logging.getLogger(__name__)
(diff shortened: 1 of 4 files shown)


intranet/apps/signage/views.py
```diff
@@ -26,3 +31,95 @@ def signage_display(request, display_id):
...
+
+    if block_id is None:
+        block = EighthBlock.objects.get_first_upcoming_block()
+        if block is None:
+            last_block = EighthBlock.objects.order_by("date").last()
+            if last_block is not None:
+                block = last_block
+    else:
+        try:
+            block = (EighthBlock.objects.prefetch_related("eighthscheduledactivity_set").get(id=block_id))
```
theo-o: Why these parentheses?
Do we want to constrain getting a block from this school year only?

intranet/apps/signage/views.py
```diff
@@ -26,3 +31,95 @@ def signage_display(request, display_id):
...
+        except EighthBlock.DoesNotExist:
+            if EighthBlock.objects.count() == 0:
```
theo-o: Isn't this condition basically going to be never met unless we recreate all of Ion? I think this might be better if we check if the number of eighth blocks for this year is 0.
Also this probably should use `exists`

intranet/apps/signage/views.py
```diff
@@ -26,3 +31,95 @@ def signage_display(request, display_id):
...
+    use_scroll = not touch_signage and bool(request.GET.get("scroll"))
+
+    context = {
+        "user": user,
+        "real_user": request.user,
+        "block_info": block_info,
+        "activities_list": safe_json(block_info["activities"]),
+        "active_block": block,
+        "active_block_current_signup": None,
+        "no_title": bool(request.GET.get("no_title")),
```
theo-o: Probably should have a default.

intranet/apps/signage/views.py
```diff
@@ -26,3 +31,95 @@ def signage_display(request, display_id):
...
+        "no_detail": bool(request.GET.get("no_detail")),
```
theo-o: Probably should have a default.

intranet/apps/signage/views.py
```diff
@@ -26,3 +31,95 @@ def signage_display(request, display_id):
...
+        "no_rooms": bool(request.GET.get("no_rooms")),
```
theo-o: Probably should have a default.

intranet/apps/signage/views.py
```diff
@@ -26,3 +31,95 @@ def signage_display(request, display_id):
...
+        "use_scroll": use_scroll,
+        "do_reload": not bool(request.GET.get("no_reload")),
```
theo-o: Probably should have a default.

intranet/apps/signage/views.py
```diff
@@ -26,3 +31,95 @@ def signage_display(request, display_id):
     context["sign"] = sign
     context["page_args"] = (sign, request)
     return render(request, "signage/base.html", context)
+
+
+def eighth(request):
```
theo-o: We should probably have a docstring.

intranet/apps/signage/views.py
```diff
@@ -26,3 +31,95 @@ def signage_display(request, display_id):
...
+            block = (EighthBlock.objects.prefetch_related("eighthscheduledactivity_set").get(id=block_id))
+        except EighthBlock.DoesNotExist:
+            if EighthBlock.objects.count() == 0:
+                # No blocks have been added yet
+                return render(request, "eighth/display.html", {"no_blocks": True})
+            else:
+                # The provided block_id is invalid
+                raise http.Http404
+
+    if "block_increment" in request.GET:
```
theo-o: It would be nice to have a comment explaining what `block_increment` is, since it takes a bit of time to realize what it is.

intranet/apps/signage/views.py
```diff
@@ -26,3 +31,95 @@ def signage_display(request, display_id):
...
+        if prev_blocks.count() > index:
+            block = prev_blocks[index]
+
+    user = User.get_signage_user()
+
+    serializer_context = {"request": request, "user": user}
+    block_info = EighthBlockDetailSerializer(block, context=serializer_context).data
+    try:
+        reload_mins = float(request.GET.get("reload_mins") or 5)
+    except Exception:
```
theo-o: We should use more specific exception(s)

intranet/apps/signage/views.py
```diff
@@ -26,3 +31,95 @@ def signage_display(request, display_id):
...
+        reload_mins = 5
+
+    touch_signage = not request.GET.get("no_touch")
+
+    next_block = list(block.next_blocks(1))
```
theo-o: Why convert this sliced queryset to a list? You can just take the first element of it on this line + 2.

intranet/apps/signage/views.py
```diff
@@ -26,3 +31,95 @@ def signage_display(request, display_id):
...
+    if next_block:
+        next_block = next_block[0]
+        if next_block.date != block.date:
+            next_block = None
+
+    prev_block = list(block.previous_blocks(1))
```
theo-o: Same

intranet/apps/signage/views.py
```diff
@@ -26,3 +31,95 @@ def signage_display(request, display_id):
...
+    if prev_block:
+        prev_block = prev_block[0]
+        if prev_block.date != block.date:
+            prev_block = None
+
+    use_scroll = not touch_signage and bool(request.GET.get("scroll"))
+
+    context = {
+        "user": user,
+        "real_user": request.user,
```
theo-o: `real_user` is not used in the template.

intranet/apps/signage/views.py
anonymoose2: Resolved in a force-push to fix the build.

theo-o: My second question still stands. Is it the intention of this page to only lookup blocks for this year or for any year?

anonymoose2: This page will only look up the next block within this year if not passed any special parameters, but it will allow looking up blocks for any year if requested.

intranet/apps/signage/views.py
anonymoose2: Elsewhere in the codebase, we frequently use `.get(...)` without a default. Is there a specific reason to add a default here?

theo-o: Thinking about it further a default is not necessarily needed.

intranet/apps/signage/views.py
```diff
@@ -26,3 +30,87 @@ def signage_display(request, display_id):
...
+    in order to switch blocks or update information.."""
+    internal_ip = check_internal_ip(request)
+    if internal_ip:
+        return internal_ip
+
+    block_id = request.GET.get("block_id")
+    if block_id is None:
+        block = EighthBlock.objects.get_first_upcoming_block()
+        if block is None:
+            block = EighthBlock.objects.order_by("date").last()
```
theo-o: This is block is not necessarily during this year.

PR #902, opened 2019-11-16
Use Page.order to sort signage pages

anonymoose2:
## Proposed changes
- Use Page.order to sort signage pages

While attempting to deploy the eighth period signage page (added in 4f4c31599a549840666dc3798d4cc0adedfd46be) in production, I tried to set the `order` on the eighth page such that it would be placed above certain other pages. However, Ion did not recognize this option, preventing me from placing it correctly.

### DIFF
--- a/intranet/apps/signage/models.py
+++ b/intranet/apps/signage/models.py
@@ -1,6 +1,20 @@
 from django.db import models
 
 
+class PageQuerySet(models.query.QuerySet):
+    def order_properly(self) -> "models.query.QuerySet[Page]":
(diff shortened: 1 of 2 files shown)


intranet/apps/signage/models.py
```diff
@@ -1,6 +1,11 @@
 from django.db import models
 
 
+class PageQuerySet(models.query.QuerySet):
+    def order_properly(self):
```
theo-o: Maybe a docstring?

PR #903, opened 2019-11-17
Add link to session management page and create a "feature announcement" for it

anonymoose2:
## Proposed changes
- Move "trust session" logic into helper function
- Add link to session management page in top nav dropdown
- Add "Trust this device" checkbox to login page
- Add feature announcement for session management page

The session management has been tested both in development environments and in production. We should announce it and make it easy for users to use.


intranet/apps/sessionmgmt/helpers.py
```diff
@@ -0,0 +1,37 @@
+from .models import TrustedSession
+
+
+def trust_session(request) -> None:
```
theo-o: Should probably have a docstring.

PR #914, opened 2019-11-23
Fix font issues in bus display on signage

anonymoose2:
## Proposed changes
- Fix font issues in bus display on signage

In production, the signage displays don't have Helvetica and Arial (the default fonts specified by SVG.js) installed. Instead, they fall back on the `sans-serif` default. Whatever this defaults to is a significantly different size from Helvetica and Arial, which results in behavior such as the route names overflowing the bus rectangles.

This change also allows adjusting the font size to something more legible.

### DIFF
--- a/intranet/static/js/bus.js
+++ b/intranet/static/js/bus.js
@@ -335,6 +335,9 @@ $(function() {
                         text.path(space.getAttribute('d'));
                         text.textPath().attr("path", space.getAttribute('d'));
                         text.style('pointer-events', 'none');
+                        // Signage displays may not have Helvetica or Arial installed, so we provide some sane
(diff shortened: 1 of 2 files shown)


intranet/static/js/bus.js
```diff
@@ -335,6 +335,7 @@ $(function() {
                         text.path(space.getAttribute('d'));
                         text.textPath().attr("path", space.getAttribute('d'));
                         text.style('pointer-events', 'none');
+                        text.font("family", "Helvetica, Arial, 'Open Sans', 'Liberation Sans', sans-serif");
```
theo-o: Maybe a comment explaining why this is is necessary to future readers?

PR #927, opened 2019-12-15
Emergency app updates

### LINKED ISSUE #926: Use celery for emergency page update
## Feature description
As described in [the Celery docs](https://docs.celeryproject.org/en/latest/userguide/periodic-tasks.html), we can implement Celery Beat to schedule tasks to run with Celery. The [`check_emerg` code](https://github.com/tjcsl/ion/blob/0c08779bdc12d94f7af70291e023708d55124667/intranet/apps/emerg/views.py#L14) is currently run every load of the index view (after checking the cache). We should use Celery beat to run `check_emerg`.

The current means of obtaining emergency info is non-optimal.

- Celery beat task that periodically updates emergency info that is returned

anonymoose2:
## Proposed changes
- Check for "under maintenance" message in emergency info
- Update FCPS emergency info in a Celery task (closes #926)

- We should not show messages to the user if FCPS's website goes into maintenance mode.
- As explained in #926, the current strategy of updating emergency announcements in the views is suboptimal.

### DIFF
--- a/Ion.egg-info/SOURCES.txt
+++ b/Ion.egg-info/SOURCES.txt
@@ -336,6 +336,7 @@ intranet/apps/emailfwd/migrations/0001_initial.py
 intranet/apps/emailfwd/migrations/__init__.py
 intranet/apps/emerg/__init__.py
 intranet/apps/emerg/api.py
+intranet/apps/emerg/tasks.py
 intranet/apps/emerg/views.py
 intranet/apps/emerg/migrations/__init__.py
 intranet/apps/error/__init__.py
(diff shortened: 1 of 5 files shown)


anonymoose2:
Note: I will handle the necessary changes to production Ion.

anonymoose2:
> Why do we still call `get_emerg_result`?

As mentioned in the commit message, I want the emergency announcement text to still update even if Celery stops working, so I kept all the existing logic in place. It just won't run unless something's broken.

theo-o:
Could we have a code comment explaining that?

PR #931, opened 2019-12-22
Locking changes

anonymoose2:
## Proposed changes
- Add generalized locking system (replaces case-speciifc system in cc257a57a977bae9e9a77876d26ece9ed94819f7)
- Lock on the user when toggling favorited activities

The locking code added in cc257a57a977bae9e9a77876d26ece9ed94819f7, if locking code was added elsewhere, could end up creating deadlock situations.

The `lock_on()` method added here attempts to avoid deadlocks by sorting the tables to lock on in a consistent way, ensuring that each thread would try to lock on the objects in the same order and avoiding deadlocks..

With regards to adding locking code to the favorited activities code, in production, there have been a lot of `IntegrityError`s emitted by this view when it tries to save and there's a duplicate object because of concurrent access. Since it already runs in a transaction, adding locking is easy.

### DIFF
--- a/intranet/apps/eighth/models.py
+++ b/intranet/apps/eighth/models.py
@@ -20,6 +20,7 @@ from django.utils import formats, timezone
 
 from ...utils.date import get_date_range_this_year, is_current_year
 from ...utils.deletion import set_historical_user
+from ...utils.locking import lock_on
 from ..notifications.tasks import email_send_task
 from . import exceptions as eighth_exceptions
 
@@ -1276,16 +1277,10 @@ class EighthScheduledActivity(AbstractBaseEighthModel):
             all_sched_act.append(sibling)
             all_blocks.append(sibling.block)
 
-        # Lock on the User and the EighthScheduledActivity. These are dummy queries that just exist to acquire the lock.
-        # You might be tempted to make these into `assert`s, but don't. That will disable locking when optimizations are enabled (enabling
-        # optimizations disables `assert`s).
-        # See https://docs.djangoproject.com/en/2.2/ref/models/querysets/#select-for-update
-        logging.debug("Locking on user %d", user.id)
-        _ = get_user_model().objects.filter(id=user.id).select_for_update().exists()
-        logging.debug("Successfully locked on user %d", user.id)
-        logging.debug("Locking on scheduled activity %d", self.id)
-        _ = EighthScheduledActivity.objects.filter(id=self.id).select_for_update().exists()
-        logging.debug("Successfully locked on scheduled activity %d", self.id)
+        # Lock on the User and the EighthScheduledActivity to prevent duplicates.
+        logger.debug("Locking on user %d and scheduled activity %d", user.id, self.id)
+        lock_on([user, self])
+        logger.debug("Successfully locked on user %d and scheduled activity %d", user.id, self.id)
 
         waitlist = None
         if force:
--- a/intranet/apps/eighth/views/signup.py
+++ b/intranet/apps/eighth/views/signup.py
@@ -14,6 +14,7 @@ from django.utils import timezone
 from django.views.decorators.http import require_POST
 
 from ....utils.helpers import is_entirely_digit
+from ....utils.locking import lock_on
 from ....utils.serialization import safe_json
 from ...auth.decorators import deny_restricted, eighth_admin_required
 from ..exceptions import SignupException
@@ -339,6 +340,8 @@ def toggle_favorite_view(request):
     aid = request.POST["aid"]
     with transaction.atomic():
         activity = get_object_or_404(EighthActivity, id=aid)
+        # Lock on the User to prevent duplicates.
+        lock_on([request.user])
         if activity.favorites.filter(id=request.user.id).nocache().exists():
             activity.favorites.remove(request.user)
             return http.HttpResponse("Unfavorited activity.")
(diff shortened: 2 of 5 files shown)


intranet/apps/eighth/models.py
```diff
@@ -1276,16 +1277,10 @@ def add_user(
...
-        # You might be tempted to make these into `assert`s, but don't. That will disable locking when optimizations are enabled (enabling
-        # optimizations disables `assert`s).
-        # See https://docs.djangoproject.com/en/2.2/ref/models/querysets/#select-for-update
-        logging.debug("Locking on user %d", user.id)
-        _ = get_user_model().objects.filter(id=user.id).select_for_update().exists()
-        logging.debug("Successfully locked on user %d", user.id)
-        logging.debug("Locking on scheduled activity %d", self.id)
-        _ = EighthScheduledActivity.objects.filter(id=self.id).select_for_update().exists()
-        logging.debug("Successfully locked on scheduled activity %d", self.id)
+        # Lock on the User and the EighthScheduledActivity.
```
theo-o:
```suggestion
        # Lock on the User and the EighthScheduledActivity to prevent duplicates.
```

intranet/utils/locking.py
```diff
@@ -0,0 +1,72 @@
...
+
+def lock_on(items: Iterable[Union[Model, Manager, QuerySet]]) -> None:
+    """Given an iterable of ``Model`` instances, ``Manager``s, and/or ``QuerySet``s, locks the
+    corresponding database rows.
+
+    More specifically, this uses the Django ORM's ``select_for_update()`` method, which translates
+    to a ``SELECT FOR UPDATE`` SQL query. For more information on what this actually does in
+    PostgreSQL (used as the database backend in dev environments and in production) see PostgreSQL's
+    documentation on locking at
+    https://www.postgresql.org/docs/8.3/explicit-locking.html#LOCKING-ROWS.
```
theo-o:
```suggestion
    https://www.postgresql.org/docs/current/explicit-locking.html#LOCKING-ROWS.
```
We should not necessarily pin documentation to a specific PostgreSQL version.

intranet/utils/locking.py
```diff
@@ -0,0 +1,72 @@
...
+
+        # Get the model class and the PKs
+        model_class = objects[0].__class__
+        object_pks = [obj.pk for obj in objects]
+
+        # Now construct a QuerySet and add it to the list
+        querysets_by_model.setdefault(model_fullname, [])
+        querysets_by_model[model_fullname].append(model_class.objects.filter(pk__in=object_pks))
+
+    # Now, we actually do the locking.
```
theo-o:
```suggestion
    # Now, with all lists of QuerySets to lock in querysets_by_model, we actually do the locking.
```

intranet/utils/locking.py
```diff
@@ -0,0 +1,72 @@
...
+from django.db.models import Manager, Model, QuerySet
+
+
+def lock_on(items: Iterable[Union[Model, Manager, QuerySet]]) -> None:
+    """Given an iterable of ``Model`` instances, ``Manager``s, and/or ``QuerySet``s, locks the
+    corresponding database rows.
+
+    More specifically, this uses the Django ORM's ``select_for_update()`` method, which translates
+    to a ``SELECT FOR UPDATE`` SQL query. For more information on what this actually does in
+    PostgreSQL (used as the database backend in dev environments and in production) see PostgreSQL's
```
theo-o:
```suggestion
    PostgreSQL (used as the database backend in all environments) see PostgreSQL's
```

intranet/apps/eighth/views/signup.py
```diff
@@ -339,6 +340,7 @@ def toggle_favorite_view(request):
     aid = request.POST["aid"]
     with transaction.atomic():
         activity = get_object_or_404(EighthActivity, id=aid)
+        lock_on([request.user])
```
theo-o:
```suggestion
        # Lock on the User to prevent duplicates.
        lock_on([request.user])
```

intranet/utils/locking.py
```diff
@@ -0,0 +1,72 @@
...
+    documentation on locking at
+    https://www.postgresql.org/docs/8.3/explicit-locking.html#LOCKING-ROWS.
+
+    This MUST by run in a transaction. A straightforward way to do this is to use the
```
theo-o:
```suggestion
    As described in Django's documentation at
    https://docs.djangoproject.com/en/stable/ref/models/queryset/#django.db.models.query.QuerySet.select_for_update, 
    the ``select_for_update`` locks prevent other transactions from acquiring locks until
    this transaction is complete.

    This MUST by run in a transaction. A straightforward way to do this is to use the
```

intranet/utils/locking.py
```diff
@@ -0,0 +1,72 @@
...
+        querysets_by_model.setdefault(model_fullname, [])
+        querysets_by_model[model_fullname].append(model_class.objects.filter(pk__in=object_pks))
+
+    # Now, we actually do the locking.
+    # We sort by the dotted name to the model class in an attempt to prevent deadlocks. If all workers
+    # attempt to lock rows in the same order, they should never deadlock.
+    for model_fullname in sorted(querysets_by_model.keys()):
+        qs_list = querysets_by_model[model_fullname]
+
+        assert qs_list
```
theo-o:
```suggestion
        # If this assertion fails, the above code has been modified in a way that breaks it
        assert qs_list
```

PR #939, opened 2020-01-11
Miscellaneous fixes

anonymoose2:
## Proposed changes
- Move administrative activities to the end of the list on signup pages
- Darken text on dashboard/login warnings in dark mode

Both are minor issues.


intranet/apps/eighth/serializers.py
```diff
@@ -8,7 +8,7 @@
 
 from django.contrib.auth import get_user_model
 from django.db import transaction
-from django.db.models import Count
+from django.db.models import Count, Q
```
theo-o: Why import of `Q`?

anonymoose2: I needed it in some earlier versions of this code and forgot to remove it. Now removed since it broke the build.

Issue #876, opened 2019-11-02
Warn about effects of cancelling sticky activities


anonymoose2:
## Feature description
When an eighth period administrator is cancelling a sticky activity, they should be warned that this will allow students to switch out of it.

## Rationale for adding feature
ea2d6affc6dcdb1df7b3faa2204bd28a798fa26e allows students to switch out of sticky activityies. However, eighth period administrators may not know this and instead assume that students are still stickied in, perhaps doing things such as uncancelling the activity and assuming all the users are still signed up.

## Suggested implementation(s) (if applicable)
-

PR #940, opened 2020-01-12
Warn about effects of cancelling sticky activities.

DasBundesland:
## Proposed changes
- Ask for confirmation before an admin cancels a sticky activity.

To close issue #876.

### DIFF
--- a/intranet/static/js/eighth/schedule.js
+++ b/intranet/static/js/eighth/schedule.js
@@ -1,4 +1,5 @@
 /* global $ */
+/* global activitySticky */
 $(function() {
     var toggleRow = function(tr) {
         var i = 0;
@@ -173,6 +174,19 @@ $(function() {
             }
         });
         if (activities !== "" && !confirm("Are you sure you want to add the following activities without a sponsor?\n" + activities)) {
+            e.preventDefault();
+        }
+        var stickyCancel = "";
+        $("tr.form-row.scheduled").each(function() {
+            var scheduledField = $(this).find("td[data-field='scheduled'] input[type='checkbox']");
+            var stickyField = $(this).find("td[data-field='sticky'] input[type='checkbox']");
+            if (!scheduledField.prop("checked")) {
+                if(stickyField.prop("checked") || activitySticky) {
+                    stickyCancel += "\n    " + $(this).find(".block-name a.ui-link").text().trim();
+                }
+            }
+        });
+        if (stickyCancel !== "" && !confirm("Cancelling the following 'sticky' activities will allow students to switch out of them. Are you sure?\n" + stickyCancel)){
             e.preventDefault();
         }
     });
--- a/intranet/templates/eighth/admin/schedule_activity.html
+++ b/intranet/templates/eighth/admin/schedule_activity.html
@@ -14,6 +14,7 @@
         window.all_sponsors = _.values($.parseJSON("{{ sponsors_json }}"));
         window.all_rooms_lookup = $.parseJSON("{{ rooms_json }}");
         window.all_rooms = _.values(all_rooms_lookup);
+        window.activitySticky = {{ activity.sticky | yesno:"true,false" }};
     </script>
 
     <script>


DasBundesland:
Done.

intranet/static/js/eighth/schedule.js
```diff
@@ -175,5 +175,8 @@ $(function() {
         if (activities !== "" && !confirm("Are you sure you want to add the following activities without a sponsor?\n" + activities)) {
             e.preventDefault();
         }
+        if($("input[type='checkbox'].unschedule").prop("checked") && $("input[type=checkbox].sticky").prop("checked") > 0 && !confirm("Cancelling a sticky activity allows students to switch out of it. \nDo you want to continue?")){
```
anonymoose2: First, instead of checking for the `unschedule` checkbox being checked, you should check for the `scheduled` box *not* being checked (but only in a `<tr>` that has the `scheduled` class).

This isn't obvious unless you've used the interface before, but the "unschedule" box is used to completely *delete* a scheduling of an activity. In that case, this dialog does not need to be shown, as all the signup objects will be deleted and users will need to sign up for other activities. *Cancelling* a scheduling of an activity is done by unchecking the main "schedule" box.

Also, even with that change implemented, this current logic will lead to the following undesired behaviors:
1. A confirmation dialog will be shown if *any* scheduling of an activity is being cancelled as long as *any* scheduling of that activity is sticky, even if these flags are set on different schedulings of the activity. You may need to implement logic similar to the `each` loop a few lines up.
2. A confirmation dialog will not be shown if the sticky flag is set on the *activity* (which implicitly applies to all activities). You may need to modify the template to export the `activity.sticky` flag as a Javascript variable in order to easily check for this.

DasBundesland: Will do.

intranet/static/js/eighth/schedule.js
```diff
@@ -175,5 +175,20 @@ $(function() {
...
             e.preventDefault();
         }
+        $("tr.form-row:not(.hidden)").each(function(i, el) {
+          if($(this).hasClass("scheduled")){
+              var stickyCancel = "";
+              var scheduledField = $("td[data-field='scheduled'] input", el);
+              var stickyField = $("td[data-field='sticky'] input", el);
+              if (!scheduledField.prop("checked")) {
+                  // TODO: Implement check for sticky flag on activity.
+                  if(stickyField.prop("checked") || ((activityIsSticky == True) && !$("input[type='checkbox'].sticky").prop("checked"))) {
```
anonymoose2:
```suggestion
                  if(stickyField.prop("checked") || activityIsSticky) {
```
I *think* this is the right condition. Regardless, your use of `True` is incorrect -- in Javascript it's `true`.

intranet/templates/eighth/admin/schedule_activity.html
```diff
@@ -18,6 +18,7 @@
 
     <script>
     $(function() {
+        var activityIsSticky = "{{ activity.sticky }}";
```
anonymoose2: This won't do what you want.

1. You need to move this out of the initialization function. Currently, it's scoped locally to that function, and the other code can't see it.
2. You're assigning this variable the *string* `"True"` or `"False"`; your comparison in `static/js/eighth/schedule.js` will always fail. Instead, try taking a look at the setting of `window.isAdmin` in `templates/signage/pages/bus.html`. That's a working of setting a Javascript boolean variable in a Django template.

intranet/static/js/eighth/schedule.js
```diff
@@ -149,7 +149,7 @@ $(function() {
                     var ntd = $("td[data-field='" + field + "']", $(this));
                     var ninp = $("input, select, textarea", ntd);
                     ninp.val(modVal);
-                });
+                });(
```
anonymoose2:
```suggestion
                });
```

intranet/static/js/eighth/schedule.js
```diff
@@ -175,5 +175,20 @@ $(function() {
         if (activities !== "" && !confirm("Are you sure you want to add the following activities without a sponsor?\n" + activities)) {
             e.preventDefault();
         }
+        $("tr.form-row:not(.hidden)").each(function(i, el) {
+          if($(this).hasClass("scheduled")){
+              var stickyCancel = "";
```
anonymoose2: You need to move this outside of this callback function so it's visible in the `confirm()` code.

intranet/templates/eighth/admin/schedule_activity.html
DasBundesland: Yeah I was pretty sure this was wrong, not super familiar with Django templates.

anonymoose2:
Looks pretty good so far. Can you [create development environment](https://github.com/tjcsl/ion/blob/master/docs/setup/vagrant.rst) and test this out?

intranet/static/js/eighth/schedule.js
```diff
@@ -175,5 +175,21 @@ $(function() {
         if (activities !== "" && !confirm("Are you sure you want to add the following activities without a sponsor?\n" + activities)) {
             e.preventDefault();
         }
+        var stickyCancel = "";
+
+        $("tr.form-row:not(.hidden)").each(function(i, el) {
+          if($(this).hasClass("scheduled")){
```
anonymoose2:
```suggestion
        $("tr.form-row.scheduled:not(.hidden)").each(function() {
```

(Assuming the corresponding changes to the rest of the function, of course.)

intranet/static/js/eighth/schedule.js
```diff
@@ -175,5 +175,21 @@ $(function() {
...
+
+              var scheduledField = $("td[data-field='scheduled'] input", el);
+              var stickyField = $("td[data-field='sticky'] input", el);
+              if (!scheduledField.prop("checked")) {
+                  if(stickyField.prop("checked") || activity_sticky) {
+                      stickyCancel += "\n    " + $(".block-name a.ui-link", el).text().trim();
```
anonymoose2:
```suggestion
                      stickyCancel += "\n    " + $(this).find(".block-name a.ui-link").text().trim();
```

See above.

intranet/static/js/eighth/schedule.js
```diff
@@ -175,5 +175,21 @@ $(function() {
...
+                  }
+              }
+          }
+      });
+      if (stickyCancel !== "" && !confirm("Cancelling the following sticky activities will allow students to switch out of them. Are you sure?\n" + stickyCancel)){
+          e.preventDefault();
+      }
     });
-});
```
anonymoose2: Please add this line back. It breaks the JS. If you tested this code, this change should have broken it

theo-o:
Hi @DasBundesland, have you been able to get around to the above requested changes?

DasBundesland:
I forgot about this and I apologize profusely. Will finish this soon.

anonymoose2:
Commit message formatting is now enforced in CI. To fix the build, you must reformat your commit message. See [`scripts/validate-commit-messages.py`](https://github.com/tjcsl/ion/blob/master/scripts/validate-commit-messages.py) for the details and previous commits for examples.

anonymoose2:
Please change the body of your commit to `Closes #876` so it will automatically close issue #876 when merged.

anonymoose2:
You have not changed the body of your commit message properly. Please do so.

intranet/static/js/eighth/schedule.js
```diff
@@ -175,5 +176,18 @@ $(function() {
         if (activities !== "" && !confirm("Are you sure you want to add the following activities without a sponsor?\n" + activities)) {
             e.preventDefault();
         }
+        var stickyCancel = "";
+        $("tr.form-row.scheduled:not(.hidden)").each(function() {
```
anonymoose2, line 180:
```suggestion
        $("tr.form-row.scheduled").each(function() {
```

This proved necessary during testing.

intranet/static/js/eighth/schedule.js
```diff
@@ -175,5 +176,18 @@ $(function() {
...
+        $("tr.form-row.scheduled").each(function() {
+            var scheduledField = $(this).find("td[data-field='scheduled'] input[type='checkbox']");
+            var stickyField = $(this).find("td[data-field='sticky'] input[type='checkbox']");
+            if (!scheduledField.prop("checked")) {
+                if(stickyField.prop("checked") || activitySticky) {
+                    stickyCancel += "\n    " + $(this).find(".block-name a.ui-link").text().trim();
+                }
+            }
+        });
+        if (stickyCancel !== "" && !confirm("Cancelling the following sticky activities will allow students to switch out of them. Are you sure?\n" + stickyCancel)){
```
theo-o, line 189: I would probably add single quotes around sticky.

DasBundesland: Just confirming. Like this? `"Cancelling the following 'sticky' activities..."`

theo-o: Yes

PR #942, opened 2020-01-17
Miscellaneous fixes

### LINKED ISSUE #913: Improve access restrictions for attendance-only users
## What needs to be refactored
Improve access restrictions for attendance-only users (`user_type="user"`).

## Rationale
Attendance-only users should only be able to take attendance and access other basic parts of Ion.

Pages that attendance-only users should have access to (excluding publicly accessible pages; this list may be incomplete):
- Eighth attendance page (should not be able to view rosters for other activities; currently not enforced)
- Preferences
- *Their* profile; nobody else's (this is known not to be enforced properly)

Pages that we should discuss giving attendance-only users access to:
- Session management
- Feedback
- Dashboard

anonymoose2:
## Proposed changes
- `django.conf.urls.url` -> `django.urls.re_path` (I missed some occurrences in 2aa3483)
- Fix `User.user_with_student_id()` and `User.user_with_ion_id() `
- Set minimum length for `EighthActivity` name to 4
- Improve access restrictions for restricted users (closes #913)
- Migrate eighth models (I think the Django 2.2 upgrade is the reason this migration is needed)

- A future release of Django may deprecate `django.conf.urls.url`.
- The `User.user_with_*_id()` methods are buggy.
- Activities should have long names.
- See #913.
- I'm not sure why this migration was necessary, but it does not appear to harm anything and has potential to interfere with normal operations later if not created.


intranet/apps/announcements/models.py
```diff
@@ -24,6 +24,9 @@ def visible_to_user(self, user):
 
         """
 
+        if user.is_restricted:
```
theo-o: This should probably have a comment.

PR #950, opened 2020-01-30
Fixes

anonymoose2:
## Proposed changes
- Add `django-redis-sessions` wrapper to fix undesired behavior
- Remove unneeded `python-gssapi` dependency
- Improve temporary file handling in printing view
- Perform minor cleanup on printing code
- Display printer descriptions instead of printer names on print page
- Fix method of counting pages for text files

- Because of eccentricities in `django-redis-sessions` and Django's session middleware, if a user visits Ion while Redis is down, their session will be cleared and they will be logged out without warning. This wrapper should fix that behavior.
- `python-gssapi` is no longer needed.
- Temporary printing files:
  - Fix potential race condition caused by improper usage of `tempfile.NamedTemporaryFile`
  - Remove temporary files at the end (we still have the original upload; we just don't need redundant copies)
- The printing app honestly needs a complete rewrite, but some of this code just makes my eyes hurt too badly to wait that long.
- The raw printer names cannot contain spaces or other special characters, so they are not very user-friendly. The printer descriptions, however, can, so we should display those to the user instead.
- The current method of counting pages in text files is flawed.

### DIFF
--- a/intranet/apps/printing/views.py
+++ b/intranet/apps/printing/views.py
@@ -54,13 +54,9 @@ def get_printers() -> List[str]:
         names = []
         for line in lines:
             if "requests since" in line:
-                names.append(line.split(" ", 1)[0])
-
-        if "Please_Select_a_Printer" in names:
-            names.remove("Please_Select_a_Printer")
-
-        if "" in names:
-            names.remove("")
+                name = line.split(" ", 1)[0]
+                if name and name != "Please_Select_a_Printer":
+                    names.append(name)
 
         cache.set(key, names, timeout=settings.CACHE_AGE["printers_list"])
         return names
@@ -309,7 +305,7 @@ def print_job(obj: PrintJob, do_print: bool = True):
             )
 
         if do_print:
-            args = ["lpr", "-P", "{}".format(printer), "{}".format(final_filename)]
+            args = ["lpr", "-P", printer, final_filename]
 
             if obj.page_range:
                 args.extend(["-o", "page-ranges={}".format(obj.page_range)])
--- a/intranet/apps/printing/views.py
+++ b/intranet/apps/printing/views.py
@@ -245,84 +245,97 @@ def print_job(obj: PrintJob, do_print: bool = True):
 
     filebase = os.path.basename(fileobj.name)
     filebase_escaped = slugify(filebase)
-    tmpfile_name = tempfile.NamedTemporaryFile(prefix="ion_print_{}_{}".format(obj.user.username, filebase_escaped)).name
-    with open(tmpfile_name, "wb+") as dest:
-        for chunk in fileobj.chunks():
-            dest.write(chunk)
 
-    tmpfile_name = convert_file(tmpfile_name, filebase)
+    # The paths to files that we need to clean up at the end
+    # This needs to be a set because we may add duplicate entries to it
+    delete_filenames = set()
+    try:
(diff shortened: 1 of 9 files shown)


intranet/settings/__init__.py
```diff
@@ -446,7 +446,7 @@ def get_month_seconds():
 
 if not TESTING:
     # Settings for django-redis-sessions
```
theo-o: We should probably leave a comment here about `intranet.utils.session`

intranet/apps/printing/views.py
```diff
@@ -27,43 +27,61 @@ class InvalidInputPrintingError(Exception):
...
+                name = match.group(1)
+                if name != "Please_Select_a_Printer":
+                    # By default, use the name of the printer instead of the description
+                    printers[name] = name
+                    # Record the name of the printer so when we parse the rest of the
+                    # extended description we know which printer it's referring to.
+                    last_name = name
+            else:
+                # If we've seen a line with the name of a printer before
+                print(last_name)
```
theo-o: Is this left over from debugging? If we want to log, we should use `logging`.

intranet/apps/printing/views.py
```diff
@@ -27,43 +27,61 @@ class InvalidInputPrintingError(Exception):
...
+                if last_name is not None:
+                    match = DESCRIPTION_LINE_RE.match(line)
+                    print(line, match)
```
theo-o: Is this left over from debugging? If we want to log, we should use `logging`.

intranet/apps/printing/views.py
```diff
@@ -245,84 +263,97 @@ def print_job(obj: PrintJob, do_print: bool = True):
...
-    obj.num_pages = num_pages
-    obj.page_range = "".join(obj.page_range.split())  # remove all spaces
-    obj.save()
+            raise Exception(msg)
 
-    range_count = check_page_range(obj.page_range, obj.num_pages)
+        if get_mimetype(final_filename) == "text/plain":
+            with open(final_filename, "r") as f:
+                num_chars = sum(len(line) for line in f)
+            num_pages = num_chars // (settings.PRINTING_PAGES_LIMIT * 50 * 72)
```
theo-o: This isn't related or blocking, but do you know the reasoning behind 50 and 72?

intranet/utils/session.py
```diff
@@ -0,0 +1,29 @@
+from redis_sessions.session import SessionStore as BaseSessionStore
+
+
+class SessionStore(BaseSessionStore):  # pylint: disable=abstract-method
```
theo-o: Given that this affects authentication and custom tinkering with our authentication backend, I would recommend some sort of unit tests.

intranet/apps/printing/views.py
anonymoose2: Yes, this (and the other `print` statement a few lines down) are left over from debugging. I don't know how I missed them.

intranet/apps/printing/views.py
anonymoose2: I'm guessing 72 characters, 50 lines a page?

While I'm at it, I'll do some refactoring to make this 1) detect line wrapping and 2) just go off a single "lines per page" constant.

intranet/utils/session.py
anonymoose2: Unfortunately, I don't think that's possible, especially given that in `settings/__init__.py` we don't set `SESSION_ENGINE` during tests. We may just have to make sure to carefully explain any changes in comments.

Issue #578, opened 2018-10-05
Allow sponsors to email the students who are signed up for a club


ovkulkarni:
This will allow teachers to contact students if there is any change in plans or there is something specific that students need to bring to the activity

PR #984, opened 2020-04-11
Allow sponsors to email students signed up for a club

anonymoose2:
## Proposed changes
- Allow sponsors to email students signed up for a club

See #578

### DIFF
--- a/intranet/apps/eighth/views/attendance.py
+++ b/intranet/apps/eighth/views/attendance.py
@@ -18,6 +18,7 @@ from django.contrib import messages
 from django.contrib.auth import get_user_model
 from django.contrib.auth.decorators import login_required
 from django.db.models import Q
+from django.http import Http404
 from django.shortcuts import get_object_or_404, redirect, render
 from django.urls import reverse
 from django.utils import timezone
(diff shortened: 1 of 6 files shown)


intranet/apps/eighth/views/attendance.py
```diff
@@ -678,3 +680,41 @@ def sponsor_schedule_widget_view(request):
...
+        emails = [signup.user.notification_email for signup in scheduled_activity.eighthsignup_set.all()]
+
+        subject = settings.EMAIL_SUBJECT_PREFIX + "{}: Message from {}".format(scheduled_activity, request.user.full_name)
+        if request.POST.get("subject"):
+            subject += ": " + request.POST["subject"]
+
+        body_text = """{} has requested to send you the following email regarding {}:
+
+{}""".format(
+            request.user.full_name, scheduled_activity.activity, request.POST["body"]
```
theo-o, line 703: I don't think we necessarily want to trust the user-provided content, even when we are not rendering on the website specifically. Maybe we should apply `safe_html`?

intranet/apps/eighth/views/attendance.py
```diff
@@ -678,3 +680,41 @@ def sponsor_schedule_widget_view(request):
...
+        )
+
+        msg = EmailMessage(
+            subject=subject, body=body_text, from_email=settings.EMAIL_FROM, to=[], reply_to=[request.user.notification_email], bcc=emails,
+        )
+
+        if settings.PRODUCTION or settings.FORCE_EMAIL_SEND:
+            msg.send()
```
theo-o, line 711: Elsewhere, we use Celery tasks to send email. I think we should do this here, especially considering some activities sponsors want to send to have large capacities.

intranet/apps/eighth/views/attendance.py
anonymoose2: This is an `EmailMessage`, not an `EmailMultiAlternatives`. The default mime type set is `text/plain`; mail clients should not be parsing it as HTML.

PR #985, opened 2020-04-15
Fixes

anonymoose2:
## Proposed changes
- Allow wrapping in schedule widget block names
- Block non-superusers from editing `admin_*` groups

- The current behavior breaks the widget horribly when long names are
entered.
- Only superusers should be able to grant admin permissions.

### DIFF
--- a/intranet/apps/eighth/views/admin/groups.py
+++ b/intranet/apps/eighth/views/admin/groups.py
@@ -32,6 +32,10 @@ def add_group_view(request):
     if request.method == "POST":
         form = QuickGroupForm(request.POST)
         if form.is_valid():
+            if not request.user.can_manage_group(form.cleaned_data["name"]):
+                messages.error(request, "You must be a superuser on Ion to manage administrative groups")
+                return redirect("eighth_admin_dashboard")
+
             group = form.save()
             messages.success(request, "Successfully added group.")
             return redirect("eighth_admin_edit_group", group_id=group.id)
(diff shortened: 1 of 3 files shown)


intranet/apps/eighth/views/admin/groups.py
```diff
@@ -32,6 +32,10 @@ def add_group_view(request):
     if request.method == "POST":
         form = QuickGroupForm(request.POST)
         if form.is_valid():
+            if form.cleaned_data["name"].startswith("admin_") and not request.user.is_superuser:
```
theo-o, line 35: Could you wrap these ifs into a separate helper functions like `admin_group_permission`?

anonymoose2: I'm not sure what you're asking. Is adding a `User.has_admin_group_permission` field instead of hardcoding `is_superuser` good enough?

theo-o: I am thinking of a function of the form:
`admin_group_permission(group, request.user)`

If the group starts with `admin_` and the user is a superuser, it returns true.

anonymoose2: What about `User.can_manage_group(group: Union[Group, str]) -> bool`?

theo-o: That is also acceptable.

PR #987, opened 2020-04-30
add morning bus functionality

Laur04:
The functionality is added to both signage and Ion bus page. At noon, the page switched from the morning to the afternoon bus page to allow for early openings/closing without having to make changes. The pages switch back at midnight. 

Let the code review begin.


Laur04:
Forgot the issue closing tag, i'll add that once the checks pass

PR #988, opened 2020-04-30
feat(bus): add morning bus arrivial display closes #954

### LINKED ISSUE #954: Add Morning Bus Arrival Page
## Feature description
A page where bus admins can log the arrival of buses in the morning.  They should be able to select a route and mark that it has arrived. 

This would simplify tracking buses for attendance and security purposes in the morning.

- Create a new timestamp model in the bus app
  - Each timestamp should have a foreign key to a bus route
- Make a new page where admins can mark buses as arrived and non-admins can view a list route statuses
  - each route should display as arrived or not, and if arrived show the time it arrived
 - /bus should route between the current bus location page and the bus arrival page depending on time of day (arrival before noon, location after noon)

Laur04:
The functionality is added to both signage and Ion bus page. At noon, the page switched from the morning to the afternoon bus page to allow for early openings/closing without having to make changes. The pages switch back at midnight.

Let the code review begin.

### DIFF
--- a/intranet/apps/bus/consumers.py
+++ b/intranet/apps/bus/consumers.py
@@ -2,6 +2,7 @@ from asgiref.sync import async_to_sync
 from channels.generic.websocket import JsonWebsocketConsumer
 
 from django.conf import settings
+from django.utils import timezone
 
 from .models import Route
 
@@ -33,13 +34,14 @@ class BusConsumer(JsonWebsocketConsumer):
 
         if self.user is not None and self.user.is_authenticated and self.user.is_bus_admin:
             try:
-                route = Route.objects.get(id=content["id"])
-                route.status = content["status"]
-                if route.status == "a":
-                    route.space = content["space"]
-                else:
-                    route.space = ""
-                route.save()
+                if self.within_time_range(content["time"]):
+                    route = Route.objects.get(id=content["id"])
+                    route.status = content["status"]
+                    if content["time"] == "afternoon" and route.status == "a":
+                        route.space = content["space"]
+                    else:
+                        route.space = ""
+                    route.save()
                 data = self._serialize()
                 async_to_sync(self.channel_layer.group_send)("bus", {"type": "bus.update", "data": data})
             except Exception as e:
@@ -73,3 +75,9 @@ class BusConsumer(JsonWebsocketConsumer):
 
         data["allRoutes"] = route_list
         return data
+
+    def within_time_range(self, time):
+        now_hour = timezone.localtime().hour
+        within_morning = now_hour <= settings.BUS_PAGE_CHANGEOVER_HOUR and time == "morning"
+        within_afternoon = now_hour > settings.BUS_PAGE_CHANGEOVER_HOUR and time == "afternoon"
+        return within_morning or within_afternoon
--- a/intranet/apps/bus/views.py
+++ b/intranet/apps/bus/views.py
@@ -2,15 +2,49 @@ from django.conf import settings
 from django.contrib.auth.decorators import login_required
 from django.http import Http404
 from django.shortcuts import render
+from django.utils import timezone
 
 from ..auth.decorators import deny_restricted
+from .models import Route
 
 
 @login_required
 @deny_restricted
 def home(request):
+    now = timezone.localtime()
+    if now.hour >= settings.BUS_PAGE_CHANGEOVER_HOUR:  # afternoon
+        return afternoon(request, True)
+    else:  # morning
+        return morning(request, True)
+
+
+@login_required
+@deny_restricted
+def morning(request, on_home=False):
     if not settings.ENABLE_BUS_APP:
         raise Http404("Bus app not enabled.")
     is_bus_admin = request.user.has_admin_permission("bus")
-    ctx = {"admin": is_bus_admin, "enable_bus_driver": settings.ENABLE_BUS_DRIVER}
+    routes = Route.objects.all()
+    ctx = {
+        "admin": is_bus_admin,
+        "enable_bus_driver": settings.ENABLE_BUS_DRIVER,
+        "bus_list": routes,
+        "changeover_time": settings.BUS_PAGE_CHANGEOVER_HOUR,
+        "on_home": on_home,
+    }
+    return render(request, "bus/morning.html", context=ctx)
+
+
+@login_required
+@deny_restricted
+def afternoon(request, on_home=False):
+    if not settings.ENABLE_BUS_APP:
+        raise Http404("Bus app not enabled.")
+    is_bus_admin = request.user.has_admin_permission("bus")
+    ctx = {
+        "admin": is_bus_admin,
+        "enable_bus_driver": settings.ENABLE_BUS_DRIVER,
+        "changeover_time": settings.BUS_PAGE_CHANGEOVER_HOUR,
+        "on_home": on_home,
+    }
     return render(request, "bus/home.html", context=ctx)
--- a/intranet/apps/signage/pages.py
+++ b/intranet/apps/signage/pages.py
@@ -1,6 +1,7 @@
 # This file contains logic for server-side rendered pages
 import datetime
 
+from django.conf import settings
 from django.utils import timezone
 
 from ..announcements.models import Announcement
@@ -17,6 +18,9 @@ def announcements(page, sign, request):  # pylint: disable=unused-argument
 
 def bus(page, sign, request):  # pylint: disable=unused-argument
     now = timezone.localtime()
+    time = "afternoon"
+    if now.hour < settings.BUS_PAGE_CHANGEOVER_HOUR:
+        time = "morning"
     day = Day.objects.today()
     if day is not None and day.end_time is not None:
         end_of_day = day.end_time.date_obj(now.date())
@@ -29,4 +33,6 @@ def bus(page, sign, request):  # pylint: disable=unused-argument
         "ws_host": request.get_host(),
         "school_end_hour": end_of_day.hour,
         "school_end_time": end_of_day.minute,
+        "time": time,
+        "changeover_time": settings.BUS_PAGE_CHANGEOVER_HOUR,
     }
--- a/intranet/settings/__init__.py
+++ b/intranet/settings/__init__.py
@@ -769,6 +769,9 @@ if TJSTAR_MAP is None:
 
 SIMILAR_THRESHOLD = 5
 
+# Time that the bus page should change from morning to afternoon display
+BUS_PAGE_CHANGEOVER_HOUR = 12
+
 # Substrings of user agents to not log in the Ion access logs
 NONLOGGABLE_USER_AGENT_SUBSTRINGS = ["Prometheus", "GoogleBot", "UptimeRobot"]
 
@@ -784,7 +787,12 @@ CELERY_BEAT_SCHEDULE = {
         "schedule": FCPS_EMERGENCY_CACHE_UPDATE_INTERVAL,
         "args": (),
     },
-    "reset-routes": {"task": "intranet.apps.bus.tasks.reset_routes", "schedule": celery.schedules.crontab(hour=8, minute=0), "args": ()},
+    "reset-routes-afternoon": {"task": "intranet.apps.bus.tasks.reset_routes", "schedule": celery.schedules.crontab(hour=0, minute=0), "args": ()},
+    "reset-routes-morning": {
+        "task": "intranet.apps.bus.tasks.reset_routes",
+        "schedule": celery.schedules.crontab(hour=BUS_PAGE_CHANGEOVER_HOUR, minute=1),
+        "args": (),
+    },
 }
 
 MAINTENANCE_MODE = False
--- a/intranet/static/css/bus.scss
+++ b/intranet/static/css/bus.scss
@@ -145,3 +145,47 @@ span.bus-number {
         display: initial;
     }
 }
+
+h2.welcome-text {
+    font-size: 30px;
+}
+
+h2.title {
+    font-size: 20px;
+    padding-top: 20px;
+    display: inline-block;
+}
+
+#buses {
+    width: 125px;
+    height: 35px;
+    font-size: 20px;
+    margin: 10px 20px;
+}
+
+.update-button {
+    display: block;
+
+    button {
+        display: inline-block;
+    }
+}
+
+.list-bus {
+    column-count: 3;
+    vertical-align: top;
+
+    .display-column {
+        column-span: 1;
+        text-align: center;
+        break-inside: avoid;
+    }
+
+    .title {
+        font-size: 22px;
+    }
+
+    .list {
+        font-size: 17px;
+    }
+}
--- a/intranet/static/js/bus-shared.js
+++ b/intranet/static/js/bus-shared.js
@@ -1,0 +1,108 @@
+export function getSocket(base_url, location, document, window, time) {
+    const protocol = (location.protocol.indexOf('s') > -1) ? 'wss' : 'ws';
+    let socket;
+    if (base_url !== '') {
+        socket = new ReconnectingWebSocket(`${protocol}://${base_url}/bus/`);
+    } else {
+        socket = new ReconnectingWebSocket(`${websocketProtocol}://${websocketHost}/bus/`);
+    }
+    socket.automaticOpen = true;
+    socket.reconnectInterval = 2000;
+    socket.maxReconnectInterval = 10000;
+    socket.reconnectDecay = 1.25;
+    socket.timeoutInterval = 5000;
+    socket.maxReconnectAttempts = null;
+
+    let disconnected = false;
+    let disconnected_msg = null;
+    let keepAliveTimeoutId = null;
+
+    socket.onopen = () => {
+        if(keepAliveTimeoutId !== null) {
+            clearTimeout(keepAliveTimeoutId);
+            keepAliveTimeoutId = null;
+        }
+
+        if (disconnected_msg) {
+            disconnected_msg.update({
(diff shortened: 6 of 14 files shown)


intranet/static/js/morningbus.js
```diff
@@ -0,0 +1,114 @@
+
+$(function() {
```
theo-o, line 2: Could we eliminate redundancies between this file and `intranet/static/js/bus.js`?

intranet/apps/bus/views.py
```diff
@@ -12,5 +14,12 @@ def home(request):
     if not settings.ENABLE_BUS_APP:
         raise Http404("Bus app not enabled.")
     is_bus_admin = request.user.has_admin_permission("bus")
-    ctx = {"admin": is_bus_admin, "enable_bus_driver": settings.ENABLE_BUS_DRIVER}
-    return render(request, "bus/home.html", context=ctx)
+    now = timezone.localtime()
+    if now.hour > 12:  # afternoon bus page,
+        is_bus_admin = request.user.has_admin_permission("bus")
```
theo-o, line 19: This is unneeded, since you get the same info in line 16 (new version)

intranet/apps/signage/pages.py
```diff
@@ -17,16 +17,33 @@ def announcements(page, sign, request):  # pylint: disable=unused-argument
...
-    return {
-        "admin": False,
-        "signage": True,
-        "ws_protocol": "ws" if request.scheme == "http" else "wss",
-        "ws_host": request.get_host(),
-        "school_end_hour": end_of_day.hour,
-        "school_end_time": end_of_day.minute,
-    }
+    if now.hour < 12:  # morning bus page,
+        day = Day.objects.today()
```
theo-o, line 21: This duplication is unneeded. If you only need to change the `time` key in this dictionary, you should make a `time` variable set in a small`if`-`elif` block

intranet/templates/bus/morning.html
```diff
@@ -0,0 +1,145 @@
+{% extends "page_with_nav.html" %}
```
theo-o, line 1: Could we reduce duplication between this and `intranet/templates/bus/home.html`?

intranet/templates/signage/pages/bus.html
```diff
@@ -45,11 +45,23 @@
                 personal: "delayed. 😢"
             },
         };
+        var now = new Date();
+            var reloadTime = new Date();
```
theo-o, line 49: Lower the indentation level on the next 7 levels

intranet/templates/bus/morning.html
Laur04: I cleaned up some of the uneeded script in `intranet/templates/bus/morning.html`, but there isn't that much similarity between the two in terms of the main content, i.e. neither naturally extends the other. Most of the scripts in `{% block js %}` are the same and of course the dark mode/css imports are as well, but everything else is different. Do you want me to create a parent page for them both to extend?

theo-o: The particular redundancies I see are:
* duplicate settings in the `js` block. I would put them in an `include`d tempate
* duplicate `enable_bus_driver` logi in the `js` block, I would move them to a common template.

Laur04:
Travis CI fails if the commit message is more than one line, so I moved the closes bit back onto the first line.

anonymoose2:
> Travis CI fails if the commit message is more than one line, so I moved the closes bit back onto the first line.

That isn't quite accurate. The second line must be blank, but you can put an extended description on the third line and following. This is a fairly standard convention for Git commits.

I'll update the commit message validation to be a little clearer.

Laur04:
Not quite sure what Codacy is catching...it's complaining about my indentation, but I think it's right.

anonymoose2:
> Not quite sure what Codacy is catching...it's complaining about my indentation, but I think it's right.

We usually just rely on the build passing and ignore Codacy. In this case, I think you're right -- Codacy is just being picky about styling and we have our own conventions.

Laur04:
Okay, the build is passing and the commit messaage is fixed.

intranet/static/css/bus.scss
```diff
@@ -145,3 +145,49 @@ span.bus-number {
         display: initial;
     }
 }
+
```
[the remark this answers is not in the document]
Laur04: Yea, I'm not sure exactly how to reconstruct it, what best practices should I be following here? @anonymoose2

intranet/templates/bus/bus_base.html
```diff
@@ -0,0 +1,34 @@
+{% load static %}
```
anonymoose2, line 1: I would either a) make this a full-blown base template and have the two pages `extends` it or 2) name this file `bus_head.html`. More descriptive.

intranet/templates/signage/pages/bus.html
```diff
@@ -98,53 +106,113 @@
     svg text.extra-small {
         font-size: 90%;
     }
+
+    .col-1 {
+        display: inline-block;
+        width: 33%;
+        height: 100%;
+        vertical-align: top;
+    }
```
anonymoose2, line 115: I would avoid using common names from external libraries like Bootstrap, just in case we switch later (though it's very unlikely).

intranet/static/css/bus.scss
```diff
@@ -145,3 +145,44 @@ span.bus-number {
...
+}
+
+.admin-show {
+    width: 50%;
+    display: inline-block;
+}
+
+.list {
+    font-size: 17px;
+}
```
anonymoose2, line 188: Please add a trailing newline.

intranet/static/css/bus.scss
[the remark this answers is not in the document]
anonymoose2: To start, this is actually SCSS, not CSS. You can nest selectors and it will combine them.

That means you can organize things hierarchically, just like in the accompanying HTML. In this case, that would be very helpful -- it's not at all clear what each of these selectors is doing.

Laur04:
I reorganized the scss structure.

intranet/templates/signage/pages/bus.html
```diff
@@ -98,53 +106,113 @@
...
-        <% } %>
-    </script>
-    <script type="text/template" id="action-button-view">
-    </script>
-    <script type="text/template" id="search-widget-view">
-        <div class="search-widget">
-            <i class="back-button fas fa-arrow-left"></i><select name="buses" id="buses"></select>
+    {% if time == "morning" %}
+        <h1 class="welcome-text">Good morning!</h1>
+        <div class="wrapper-1">
```
theo-o, line 148: I would make the class names a bit more descriptive than `wrapper-1` and `wrapper-2`. Maybe `bus-heading` and `bus-body`?

intranet/templates/bus/morning.html
```diff
@@ -0,0 +1,62 @@
...
+        <h2 class="welcome-text">Good morning!</h2>
+        {% if admin %}
+            <h2 class="title">Mark bus:</h2>
+            <select id="buses">
+                {% for bus in bus_list %}
+                    <option value="{{ bus.route_name }}">{{ bus.route_name }}</option>
+                {% endfor %}
+            </select>
+            <div class="update-button">
+                <button onclick="sendMorningUpdate({'status':'a'})">Arrived</button>
```
theo-o, line 42: This feature is broken (i.e. I cannot mark buses)

Laur04: First issue there is actually related to the second...in bus.js there is an additional if statement that only makes the morning bus login work before noon. I'm going to put that into a variable in settings so the hardcoding won't be an issue in the future.

Laur04: Okay, try now. I removed all the hardcoding of the hour and configured it in settings. To debug after noon, you will have to change it to 22 or something. As a note, changing it to 24 makes celery mad, but this doesn't seem like a reasonable use case so I didn't really worry about it.

intranet/static/js/bus.js
```diff
@@ -1,707 +1,715 @@
 /* globals Messenger */
 var bus = {};
+var now = new Date();
+afterNoon = now.getHours() > parseInt("{{ changeover_time }}", 10);
```
anonymoose2, line 4: Honestly, the changes here are so large that I'd rather have them in separate files -- as it stands I can't even tell what's what to review it. Maybe keep the code to open/reopen the websocket in a `bus-shared.js`, and move everything else to `bus-morning.js` and `bus-afternoon.js`?

I also don't like the fact that `afterNoon` here is set based on the current time. Users should be able to see either the afternoon or morning bus page specifically, and the auto-switching functionality should only be on the main `/bus` URL. This would make future testing MUCH easier.

intranet/apps/bus/consumers.py
```diff
@@ -33,13 +33,18 @@ def receive_json(self, content):  # pylint: disable=arguments-differ
...
-                route = Route.objects.get(id=content["id"])
-                route.status = content["status"]
-                if route.status == "a":
-                    route.space = content["space"]
-                else:
-                    route.space = ""
-                route.save()
+                if content["time"] == "morning":
+                    route = Route.objects.get(route_name=content["id"])
+                    route.status = content["status"]
```
anonymoose2, line 38: 1. Why is `route_name` used instead of `id`? I'd rather keep the consumer the same and tweak the JS.
2. Why does the morning bus code need to be separate? Aside from the use of `route_name` instead of `id`, the only difference is that `route.space` is not set, and that's going to be reset by the Celery task anyway.

intranet/static/js/bus.js
```diff
@@ -1,707 +1,715 @@
...
 
 $(function() {
```
anonymoose2, line 6: There's another problem with this JS as it stands: Some users are probably going to open `/bus` in the morning, leave it open until the afternoon, then pull the page back up and expect it to work without refreshing. I haven't tested the UI yet, but I doubt it handles that case. And this could become important.

Here's what I would recommend: For the `home` view ONLY (this may require passing an extra variable through the `morning` and `afternoon` views), if it passes the changeover time (either way, morning -> afternoon or afternoon -> morning), then the page should display a popup that tells the user they need to refresh to display accurate results.

intranet/static/js/bus.js
Laur04: The issue I see with allowing users to see either is that they use the same bus objects, so if an admin were to make a change in the afternoon, it would make a change in the morning page that might confuse users. I think I would also need to create a separate web socket, which is going to involve lots of repeated code. What if I kept the switchover, but allowed a static version of the opposite to still be displayed. So after 12, the morning bus page could be shown, but wouldn't be connected to a web socket. This would solve everything except the refresh issue.

intranet/apps/bus/views.py
```diff
@@ -12,5 +14,16 @@ def home(request):
     if not settings.ENABLE_BUS_APP:
         raise Http404("Bus app not enabled.")
     is_bus_admin = request.user.has_admin_permission("bus")
-    ctx = {"admin": is_bus_admin, "enable_bus_driver": settings.ENABLE_BUS_DRIVER}
-    return render(request, "bus/home.html", context=ctx)
+    now = timezone.localtime()
```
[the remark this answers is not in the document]
Laur04: To make sure I'm clear, they should still both be routed under the same url pattern in urls.py, just organized into three different view functions? Or were you suggesting that there should be separate routes, maybe having the main /bus url route based on time, then a /bus/morning that always routes to the morning and and /bus/afternoon that always routes to the afternoon?

intranet/static/js/bus.js
Laur04: I force a refresh on signage, what about doing that here?

anonymoose2: I would rather leave it up to the user. @theo-o?

intranet/static/js/bus.js
anonymoose2: Ah, I missed the issue with the same bus objects. What you could do is alter the consumer so that it takes a `morning`/`afternoon` parameter in the URL, and then pretends that there are no buses if that doesn't match the actual time. The logic would be a little tricky, but I'd rather do it that way.

Also, you can't embed variables in static files like this.

intranet/settings/__init__.py
```diff
@@ -784,7 +787,12 @@ def get_log(name):  # pylint: disable=redefined-outer-name; 'name' is used as th
         "schedule": FCPS_EMERGENCY_CACHE_UPDATE_INTERVAL,
         "args": (),
     },
-    "reset-routes": {"task": "intranet.apps.bus.tasks.reset_routes", "schedule": celery.schedules.crontab(hour=8, minute=0), "args": ()},
+    "reset-routes-afternoon": {"task": "intranet.apps.bus.tasks.reset_routes", "schedule": celery.schedules.crontab(hour=8, minute=0), "args": ()},
```
anonymoose2, line 790:
```suggestion
    "reset-routes-afternoon": {"task": "intranet.apps.bus.tasks.reset_routes", "schedule": celery.schedules.crontab(hour=0, minute=0), "args": ()},
```

I'd rather reset the routes earlier. Security might pull up the page at 7:50 and be confused that the routes are still there.

And since other parts of this PR assume that it rolls over at midnight, that's a convenient time.

intranet/templates/bus/bus_head.html
```diff
@@ -0,0 +1,53 @@
...
+        var midnight = new Date();
+        midnight.setHours(24, 0, 0, 0);
+        reloadTime.setHours(parseInt("{{ changeover_time }}", 10), 00, 0);
+        if (reloadTime > now) {
+            setTimeout(function() {
+                if (confirm("This page is out of date. Reload for updated information?")) {
+                    window.location.reload(true);
+                }
+            }, reloadTime - now);
+        } else if (reloadTime <= now ) {
```
anonymoose2, line 36:
```suggestion
        } else if (reloadTime <= now) {
```

Just a minor styling nitpick.

intranet/templates/signage/pages/bus.html
```diff
@@ -98,53 +107,113 @@
...
-        <% if (collection.length === 0) { %>
-            <h3 class="empty"> <%= empty_text %> </h3>
-        <% } %>
-    </script>
-    <script type="text/template" id="action-button-view">
-    </script>
-    <script type="text/template" id="search-widget-view">
-        <div class="search-widget">
-            <i class="back-button fas fa-arrow-left"></i><select name="buses" id="buses"></select>
+    {% if time == "morning" %}
```
anonymoose2, line 147: See my earlier comments in `bus-shared.js`.

intranet/apps/bus/views.py
```diff
@@ -2,15 +2,49 @@
...
 
 from ..auth.decorators import deny_restricted
+from .models import Route
 
 
 @login_required
 @deny_restricted
 def home(request):
+    now = timezone.localtime()
+    if now.hour > settings.BUS_PAGE_CHANGEOVER_HOUR:  # afternoon
```
anonymoose2, line 15:
```suggestion
    if now.hour >= settings.BUS_PAGE_CHANGEOVER_HOUR:  # afternoon
```

This matches behavior of other pages.

intranet/static/js/bus-shared.js
```diff
@@ -0,0 +1,102 @@
...
+        }
+
+        var data = JSON.parse(event.data)
+
+        // Don't try and handle keepalives -- when handled, they effectively clear the bus board
+        if(data.type === "keepalive-response") {
+            return;
+        }
+
+        if (time === 'afternoon') {
```
anonymoose2, line 48: I think the `time` variable can be removed.

Instead, you could check for the elements that each time of day uses. If the `#arrived` element exists, you populate the morning fields. If the `#map` element exists, you send the afternoon update event.

This may seem minor, but it would allow you to do some interesting things. For example, you can render both the morning and afternoon portions on the signage page (in separate `<div>`s) and then have some JS to show/hide them based on the time. Then the page never has to refresh; it just updates dynamically.

You could also apply this to the main bus page. I know I said earlier that I'd rather have the morning and afternoon buses on separate pages, but I'd be happy with a single page if it automatically switches over so the user doesn't have to intervene.

Laur04: That makes sense for signage, but I'd like to leave the main bus page how it is because:

1. It allows Security to access whatever page they need, independent of the time. I know the websocket currently doesn't make changes to the buses outside the time frame for that page, but I could see use cases where they might ask for that restriction to be removed.
2. Like you pointed out earlier, it makes testing much, much easier in the future.
3. I think separating the code makes it easier to understand and maintain.
4. Having morning/afternoon code on the same page and switching between them with Javascript feels like it defeats the point of a django based system.

In essence, I agree on signage, but I think leaving the main bus page as it is makes it more robust. Thoughts?

anonymoose2: Hmm. @theo-o what do you think about this?

theo-o: I can see reasonable arguments for both sides of this proposal. I think it ultimately comes down to whether the benefits of having morning and afternoon portions broken out into different pages outweigh the benefits of having the portions exist on the same page.

In my opinion, @Laur04's proposal is the best (broken out on `/bus` and together on `/signage/display/[SIGNAGE]`) because:
- Bus signage is view-only, while regular bus is both read and write
- As a result, there really isn't a need for separate pages for morning and afternoon on bus signage
- Bus signage is only really applicable in the mornings and afternoons, while people may want to see both morning and afternoon on regular bus
- Security may want to see how both the morning and afternoon look like by just visiting different pages instead of waiting for the turnover

anonymoose2: Sounds good to me.

Laur04:
Okay, I rebased against dev...is there anything else?

PR #990, opened 2020-04-30
Attempt to fix build for new version of pylint

anonymoose2:
## Proposed changes
- Attempt to fix build for new version of pylint

## Brief description of rationale
A new version of Pylint has been released. This PR fixes several issues Pylint finds.

PR #991, opened 2020-05-01
Unpin pylint from 2.4.4 and fix build for 2.5.x

anonymoose2:
## Proposed changes
- Unpin pylint from 2.4.4 and fix build for 2.5.x

Once the upstream issues mentioned in #990 are resolved, we should switch back to the latest version of Pylint.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -31,7 +31,7 @@ install:
 - npm install sass
 - pip install -U pip setuptools
 - pip install -U -r requirements.txt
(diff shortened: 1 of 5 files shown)


theo-o:
According to upstream (`https://github.com/PyCQA/pylint/issues/3524#issuecomment-622933692`), the regression associated with this issue should be resolved, and a `2.5.1.` release should be forthcoming.

PR #993, opened 2020-05-11
fix(birthdays): clean up user birthday fields

Laur04:
## Proposed changes
- clean up birthday fields and methods from the users model that I had missed earlier


theo-o:
To resolve the currently-failing build on this PR, you'll need to rebase against the current `dev` to incorporate https://github.com/tjcsl/ion/commit/e15e04c1e64e55c2b17632802acaf7efa9483880

Handing over to @anonymoose2 for final review.

PR #994, opened 2020-05-11
feat(polls): add number voted count

### LINKED ISSUE #886: Add number voted/qualified voters to polls admin
## Feature description
Showing the aggregate numbers of voting users helps poll admins report to the initiating groups how well the poll is performing.

Although we don’t want admins to see results during poll open, we should expose the total number of voting users and the qualified voter count at all times. This would help to gauge the voter turnout rate.

Text showed to poll admins in the form underneath the Vote and View Results buttons.
```
X users out of X (X%) have voted on this poll
```

Laur04:
closes #886

## Proposed changes
- 
- 
-

### DIFF
--- a/intranet/apps/polls/models.py
+++ b/intranet/apps/polls/models.py
@@ -94,6 +94,21 @@ class Poll(models.Model):
                 users = list(q.get_users_voted())
         return users
 
+    def get_num_eligible_voters(self):
+        if self.groups.exists():
+            return get_user_model().objects.exclude(user_type="service").filter(groups__poll=self).distinct().count()
+        else:
+            return get_user_model().objects.exclude(user_type="service").count()
+
+    def get_percentage_voted(self, voted, able):
+        return "{:.1%}".format(0 if able == 0 else voted / able)
+
+    def get_voted_string(self):
+        users_voted = len(self.get_users_voted())
+        users_able = self.get_num_eligible_voters()
+        percent = self.get_percentage_voted(users_voted, users_able)
+        return "{} out of {} ({}) eligible users voted in this poll.".format(users_voted, users_able, percent)
+
     def has_user_voted(self, user):
         return Answer.objects.filter(question__in=self.question_set.all(), user=user).count() == self.question_set.count()
 
--- a/intranet/apps/polls/tests.py
+++ b/intranet/apps/polls/tests.py
@@ -49,6 +49,9 @@ class ApiTest(IonTestCase):
         self.assertEqual(Question.objects.filter(poll=poll[0]).count(), 1)
         self.assertEqual(Choice.objects.filter(question__poll=poll[0]).count(), 3)
 
+        # Everyone can vote, no one has
+        self.assertEqual(poll[0].get_voted_string(), "{} out of {} ({}) eligible users voted in this poll.".format(0, 1, "0.0%"))
+
         # Make sure that the user can vote in the poll
         response = self.client.post(reverse("poll_vote", kwargs={"poll_id": poll.first().id}), data={"question-1": 3})
 
@@ -56,6 +59,9 @@ class ApiTest(IonTestCase):
 
         user_choice = Answer.objects.get(question=Question.objects.get(poll=poll[0]), user=user).choice
         self.assertEqual(user_choice, Choice.objects.get(question__poll=poll[0], info="Blue"))
+
+        # Everyone can vote, one person has
+        self.assertEqual(poll[0].get_voted_string(), "{} out of {} ({}) eligible users voted in this poll.".format(1, 1, "100.0%"))
 
         # Test poll deletion
         response = self.client.post(reverse("delete_poll", kwargs={"poll_id": poll.first().id}))
@@ -77,6 +83,9 @@ class ApiTest(IonTestCase):
         future_5 = timezone.localtime() + datetime.timedelta(minutes=5)
 
         poll = Poll.objects.create(title="Test", description="Test", start_time=ago_5, end_time=ago_5, visible=False)
+
+        # Everyone can vote, no one has
+        self.assertEqual(poll.get_voted_string(), "{} out of {} ({}) eligible users voted in this poll.".format(0, 1, "0.0%"))
 
         # Happened in the past, not visible
         self.assertFalse(poll.can_vote(user))
@@ -112,9 +121,15 @@ class ApiTest(IonTestCase):
         poll.groups.add(temp_group)
         self.assertFalse(poll.can_vote(user))
 
+        # temp_group can vote, no one in temp_group
+        self.assertEqual(poll.get_voted_string(), "{} out of {} ({}) eligible users voted in this poll.".format(0, 0, "0.0%"))
+
         # Happening now, visible, in poll group
         user.groups.add(temp_group)
         self.assertTrue(poll.can_vote(user))
+
+        # temp_group can vote, one in temp_group
+        self.assertEqual(poll.get_voted_string(), "{} out of {} ({}) eligible users voted in this poll.".format(0, 1, "0.0%"))
 
         # admin_all, happened in the past, not visible
         poll.end_time = ago_5
--- a/intranet/templates/polls/poll.html
+++ b/intranet/templates/polls/poll.html
@@ -40,10 +40,15 @@
         for
         {% for group in poll.groups.all %}
             {{ group }}
-            {% if not forloop.last %},{% endif %}
+            {% if not forloop.last %},
+            {% else %}
+            {% endif %}
         {% empty %}
             everyone
         {% endfor %}
+        {% if is_polls_admin and not poll.before_start_time %}
+            &bull; {{ poll.get_voted_string }}
+        {% endif %}
     </div>
 
     {% if poll.announcement and not announcement %}
--- a/intranet/templates/polls/results.html
+++ b/intranet/templates/polls/results.html
@@ -68,6 +68,7 @@
             &nbsp; <a href="{% url 'poll_csv_results' poll.id %}" class="button small-button print-hide">Download as CSV</a>
         {% endif %}<br>
         <h2>Results: {{ poll }}</h2>
+        <p>&nbsp; {{ poll.get_voted_string }}</p>
 
         <ol class="questions">
             {% for q in questions %}


Laur04:
Build is failing on `The command "if [[ $TRAVIS_PYTHON_VERSION == 3.7 ]]; then flake8 --max-line-length 150 --exclude=*/migrations/* .; fi" exited with 1.` I'm not really sure why, it was passing before, I changed one line (making it shorter), now it fails.

theo-o:
> Build is failing on `The command "if [[ $TRAVIS_PYTHON_VERSION == 3.7 ]]; then flake8 --max-line-length 150 --exclude=*/migrations/* .; fi" exited with 1.` I'm not really sure why, it was passing before, I changed one line (making it shorter), now it fails.

This appears to be a failure in another part of the codebase. Feel free to ignore it.

intranet/templates/polls/poll.html
```diff
@@ -40,10 +40,15 @@ <h3>
...
-            {% if not forloop.last %},{% endif %}
+            {% if not forloop.last %},
+            {% else %} &bull;
+            {% endif %}
         {% empty %}
-            everyone
+            everyone &bull;
         {% endfor %}
+        {% if is_polls_admin and not poll.before_start_time %}
+            {{ poll.get_percentage_voted }} of eligible users have voted
```
theo-o, line 50: This message should be roughly similar to the message in `intranet/templates/polls/results.html` (i.e. incorporate raw numbers). I would suggest making the message a property of a `Poll` that returns the message as a string.

intranet/apps/polls/models.py
```diff
@@ -94,6 +94,23 @@ def get_users_voted(self):
...
+                else:
+                    users = list(DjangoGroup.objects.get(name=g).user_set.all())
+            return len(users)
+        else:
+            users = get_user_model().objects.all()
+            return len(users) - 1
+
+    def get_percentage_voted(self):
+        percentage = len(self.get_users_voted()) / self.get_num_eligible_voters()
+        return "{:.0%}".format(percentage)
```
theo-o, line 112: I personally think that it is nicer to have the percentage to the tenths place.

intranet/apps/polls/models.py
```diff
@@ -94,6 +94,23 @@ def get_users_voted(self):
                 users = list(q.get_users_voted())
         return users
 
+    def get_num_eligible_voters(self):
+        users = []
```
theo-o, line 98: This method makes highly inefficient queries against the database by needlessly loading possibly large querysets into memory. I do acknowledge other methods in this file (particularly `get_user_voted()`) do not serve as a good example.
I would reference the Django [docs on queries](https://docs.djangoproject.com/en/3.0/topics/db/queries/) and the Django [docs on querysets](https://docs.djangoproject.com/en/3.0/ref/models/querysets/) which are good resources when trying to construct nice and efficient DB queries

theo-o, line 98: I would make this an empty queryset with `get_user_model().objects.none()`

intranet/apps/polls/models.py
```diff
@@ -94,6 +94,23 @@ def get_users_voted(self):
...
+        if self.groups.all():
+            for g in self.groups.all():
+                if users:
+                    users = list(set(DjangoGroup.objects.get(name=g).user_set.all()) | set(users))
+                else:
+                    users = list(DjangoGroup.objects.get(name=g).user_set.all())
+            return len(users)
+        else:
+            users = get_user_model().objects.all()
```
theo-o, line 107: Use `return get_user_model().objects.count()  - 1` to avoid loading all users into memory.

intranet/apps/polls/models.py
```diff
@@ -94,6 +94,23 @@ def get_users_voted(self):
                 users = list(q.get_users_voted())
         return users
 
+    def get_num_eligible_voters(self):
+        users = []
+        if self.groups.all():
```
theo-o, line 99:
```suggestion
        if self.groups.exists():
```

`exists()` checks if the queryset exists (has a non-zero length).

intranet/apps/polls/models.py
```diff
@@ -94,6 +94,23 @@ def get_users_voted(self):
...
+            for g in self.groups.all():
+                if users:
+                    users = list(set(DjangoGroup.objects.get(name=g).user_set.all()) | set(users))
+                else:
+                    users = list(DjangoGroup.objects.get(name=g).user_set.all())
+            return len(users)
```
theo-o, line 105: I would return the `count()`  of a `users` queryset.

You could make this even more efficient with some complex queryset magic :smiley:

intranet/apps/polls/models.py
```diff
@@ -94,6 +94,23 @@ def get_users_voted(self):
                 users = list(q.get_users_voted())
         return users
 
+    def get_num_eligible_voters(self):
+        users = []
+        if self.groups.all():
+            for g in self.groups.all():
+                if users:
+                    users = list(set(DjangoGroup.objects.get(name=g).user_set.all()) | set(users))
```
theo-o, line 102: Similar behavior to this can be achieved with Django's `union`. That also eliminates the need for this `if`-`else` block.

intranet/apps/polls/models.py
Laur04: I'm concerned about doing that because users might overlap between groups. I'm trying to use Q queries to do this more efficiently.

Laur04:
Okay, I think I've made it better using Q queries.

intranet/apps/polls/models.py
```diff
@@ -94,6 +94,30 @@ def get_users_voted(self):
...
 
+    def get_num_eligible_voters(self):
+        groups = get_user_model().objects.none()
+        if self.groups.exists():
+            q_objects = Q(name__in=[])
+            for g in self.groups.all():
+                q_objects.add(Q(name=g), Q.OR)
+            groups = DjangoGroup.objects.filter(q_objects)
+            users = set()
+            for g in groups:
```
theo-o, line 105: You don't need to create sets and use the `|` operator in order to merge different querysets together without duplicates. The SQL union operator (as exposed by Django's [union()](https://docs.djangoproject.com/en/3.0/ref/models/querysets/#union)) returns a combined queryset that, by default, unless the keyword argument `all=True` is passed, only selects distinct values.

> The UNION operator selects only distinct values by default.
\- [Source](https://docs.djangoproject.com/en/3.0/ref/models/querysets/#union)

intranet/apps/polls/models.py
```diff
@@ -94,6 +94,30 @@ def get_users_voted(self):
...
+
+    def get_percentage_voted(self, voted, able):
+        percentage = voted / able
+        return "{:.1%}".format(percentage)
+
+    def get_voted_string(self):
+        users_voted = len(self.get_users_voted())
+        users_able = self.get_num_eligible_voters()
+        percent = self.get_percentage_voted(users_voted, users_able)
+        return str(users_voted) + " out of " + str(users_able) + "(" + percent + ") of eligible users voted in this poll."
```
theo-o, line 119: Especially with this many components of the string, it really makes sense to use `.format()`.

intranet/apps/polls/models.py
```diff
@@ -94,6 +94,30 @@ def get_users_voted(self):
...
+        else:
+            return get_user_model().objects.count() - 1
+
+    def get_percentage_voted(self, voted, able):
+        percentage = voted / able
+        return "{:.1%}".format(percentage)
+
+    def get_voted_string(self):
+        users_voted = len(self.get_users_voted())
+        users_able = self.get_num_eligible_voters()
```
theo-o, line 117: The edge case of eligible voters being 0 should be handled to prevent a divide-by-zero error in `get_percentage_voted`. Although it might seem like an implausible scenario, there is the possibility of a poll creator creating a group authorized to vote in a poll that eventually is cleared to zero.

intranet/apps/polls/models.py
```diff
@@ -94,6 +94,30 @@ def get_users_voted(self):
                 users = list(q.get_users_voted())
         return users
 
+    def get_num_eligible_voters(self):
+        groups = get_user_model().objects.none()
+        if self.groups.exists():
+            q_objects = Q(name__in=[])
```
theo-o, line 100: You can use just `Q()`

intranet/apps/polls/models.py
```diff
@@ -94,6 +94,32 @@ def get_users_voted(self):
...
+        groups = DjangoGroup.objects.none()
+        if self.groups.exists():
+            q_objects = Q()
+            for g in self.groups.all():
+                q_objects.add(Q(name=g), Q.OR)
+            groups = DjangoGroup.objects.filter(q_objects)
+            users = get_user_model().objects.none()
+            for g in groups:
+                users = users.union(g.user_set.all())
+            return len(users)
```
theo-o, line 107: `users.count()` is more efficient.

intranet/apps/polls/models.py
```diff
@@ -94,6 +94,32 @@ def get_users_voted(self):
...
+        else:
+            return get_user_model().objects.count() - 1
+
+    def get_percentage_voted(self, voted, able):
+        if able == 0:  # there are no eligible voters
+            return "0.0%"
```
theo-o, line 113: I would not return immediately here, and save `0.0` to percentage, so that, if the format of the returned percentage needs to change, it can be done by changing one line.

intranet/apps/polls/models.py
```diff
@@ -94,6 +94,32 @@ def get_users_voted(self):
...
+            groups = DjangoGroup.objects.filter(q_objects)
+            users = get_user_model().objects.none()
+            for g in groups:
+                users = users.union(g.user_set.all())
+            return users.count()
+        else:
+            return get_user_model().objects.count() - 1
+
+    def get_percentage_voted(self, voted, able):
+        percentage = voted / able
```
theo-o, line 112: If able is 0, this raises a divide-by-zero error. I would recommend using a `try`-`except` block.

```python
    percentage = 0
    try:
        percentage = voted / able
    except ZeroDivisionError:
        pass
```

intranet/apps/polls/models.py
```diff
@@ -94,6 +94,34 @@ def get_users_voted(self):
...
+            percentage = voted / able
+        except ZeroDivisionError:  # there are no eligible voters
+            pass
+        return "{:.1%}".format(percentage)
+
+    def get_voted_string(self):
+        users_voted = len(self.get_users_voted())
+        users_able = self.get_num_eligible_voters()
+        percent = self.get_percentage_voted(users_voted, users_able)
+        return "{} out of {} ({}) of eligible users voted in this poll.".format(str(users_voted), str(users_able), percent)
```
theo-o, line 123:
```suggestion
        return "{} out of {} ({}) eligible users voted in this poll.".format(users_voted, users_able, percent)
```

Calling `str` is unnecessary

intranet/apps/polls/models.py
```diff
@@ -94,6 +94,34 @@ def get_users_voted(self):
                 users = list(q.get_users_voted())
         return users
 
+    def get_num_eligible_voters(self):
+        groups = DjangoGroup.objects.none()
```
theo-o, line 98:
```suggestion
```

intranet/templates/polls/poll.html
```diff
@@ -40,10 +40,15 @@ <h3>
         for
         {% for group in poll.groups.all %}
             {{ group }}
-            {% if not forloop.last %},{% endif %}
```
theo-o, line 43: The bullets on line 43 & 47 should not be placed here in the logic, since, if the user is not a polls admin, there is a stray bullet at the end. The bullet should likely be placed on line 50.

intranet/templates/polls/poll.html
```diff
@@ -40,10 +40,15 @@ <h3>
...
+            {% if not forloop.last %},
+            {% else %}
+            {% endif %}
         {% empty %}
-            everyone
+            everyone
```
anonymoose2, line 47:
```suggestion
            everyone
```

intranet/templates/polls/poll.html
```diff
@@ -40,10 +40,15 @@ <h3>
...
-            {% if not forloop.last %},{% endif %}
+            {% if not forloop.last %},
+            {% else %}
+            {% endif %}
         {% empty %}
-            everyone
+            everyone 
         {% endfor %}
+        {% if is_polls_admin and not poll.before_start_time %}
+            &bull; {{ poll.get_voted_string }}
```
anonymoose2, line 50: I haven't tested the UI. Is the bullet here really necessary?

intranet/apps/polls/models.py
```diff
@@ -94,6 +94,33 @@ def get_users_voted(self):
...
+        else:
+            return get_user_model().objects.count() - 1
+
+    def get_percentage_voted(self, voted, able):
+        percentage = 0
+        try:
+            percentage = voted / able
+        except ZeroDivisionError:  # there are no eligible voters
+            pass
+        return "{:.1%}".format(percentage)
```
anonymoose2, line 116: 0 if able == 0 else voted / able
```suggestion
        return "{:.1%}".format(0 if able == 0 else voted / able)
```

intranet/apps/polls/models.py
```diff
@@ -94,6 +94,33 @@ def get_users_voted(self):
...
+            q_objects = Q()
+            for g in self.groups.all():
+                q_objects.add(Q(name=g), Q.OR)
+            groups = DjangoGroup.objects.filter(q_objects)
+            users = get_user_model().objects.none()
+            for g in groups:
+                users = users.union(g.user_set.all())
+            return users.count()
+        else:
+            return get_user_model().objects.count() - 1
```
anonymoose2, line 108: What is the meaning of `-1` here? Please add a comment explaining it for future reference.

Laur04: There is the RESET_PASSWORD user here, that obviously isn't voting in the poll, but I think is included in the count of all the users. I will comment to clarify this.

intranet/templates/polls/poll.html
Laur04: Yes, it keep the format consistent because the previous two pieces of info on that line are separated by a bullet.

intranet/apps/polls/models.py
anonymoose2: Hmm. In that case, we should instead use `.exclude(user_type="service")`. More generalized, and more obvious from the code itself what's going on.

intranet/templates/polls/poll.html
anonymoose2: Ah. Makes sense.

Laur04:
I'm not super familiar with tests in Django, so let me know if I'm missing something there.

intranet/apps/polls/tests.py
```diff
@@ -57,6 +57,9 @@ def test_poll(self):
         user_choice = Answer.objects.get(question=Question.objects.get(poll=poll[0]), user=user).choice
```
anonymoose2, line 57: These tests look good. However, can you also add a test before the `poll_vote` request above?

intranet/apps/polls/models.py
```diff
@@ -94,6 +94,28 @@ def get_users_voted(self):
...
+    def get_num_eligible_voters(self):
+        if self.groups.exists():
+            q_objects = Q()
+            for g in self.groups.all():
+                q_objects |= Q(name=g)
+            groups = DjangoGroup.objects.filter(q_objects)
+            users = get_user_model().objects.none()
+            for g in groups:
+                users = users.union(g.user_set.all())
+            return users.count()
```
anonymoose2, line 106: I know I said that was the last comment, but I think this will work (and it'll be much more efficient if it does):
```suggestion
            return get_user_model().objects.filter(groups__poll_set=self)
```

@theo-o am I getting this right? Should we test on production and make sure they give the same results?

Laur04: Shouldn't we add .count()?

anonymoose2: Yes, that would be necessary too.

theo-o:
```suggestion
    return get_user_model().objects.filter(groups__poll=self).distinct().count()
```

@anonymoose2's suggestion is not completely correct.

anonymoose2: Update: @theo-o and I experimented. The correct version is `get_user_model().objects.filter(groups__poll=self).distinct().count()`; please update.

Laur04:
Is something wrong with Travis CI again? If you look at the details, it shows all checks are passed, but it hasn't update this page.

intranet/apps/polls/models.py
```diff
@@ -94,6 +94,21 @@ def get_users_voted(self):
                 users = list(q.get_users_voted())
         return users
 
+    def get_num_eligible_voters(self):
+        if self.groups.exists():
+            return get_user_model().objects.filter(groups__poll=self).distinct().count()
```
anonymoose2, line 99:
```suggestion
            return get_user_model().objects.exclude(user_type="service").filter(groups__poll=self).distinct().count()
```

intranet/apps/polls/models.py
```diff
@@ -94,6 +94,21 @@ def get_users_voted(self):
...
+            return get_user_model().objects.exclude(user_type="service").count()
+
+    def get_percentage_voted(self, voted, able):
+        return "{:.1%}".format(0 if able == 0 else voted / able)
+
+    def get_voted_string(self):
+        users_voted = len(self.get_users_voted())
+        users_able = self.get_num_eligible_voters()
+        percent = self.get_percentage_voted(users_voted, users_able)
+        return "{} out of {} ({}) of eligible users voted in this poll.".format(users_voted, users_able, percent)
```
theo-o, line 110:
```suggestion
        return "{} out of {} ({}) eligible users voted in this poll.".format(users_voted, users_able, percent)
```

intranet/apps/polls/tests.py
```diff
@@ -49,6 +49,9 @@ def test_poll(self):
         self.assertEqual(Question.objects.filter(poll=poll[0]).count(), 1)
         self.assertEqual(Choice.objects.filter(question__poll=poll[0]).count(), 3)
 
+        # Everyone can vote, no one has
+        self.assertEqual(poll[0].get_voted_string(), "{} out of {} ({}) of eligible users voted in this poll.".format(0, 1, "0.0%"))
```
theo-o, line 53:
```suggestion
        self.assertEqual(poll[0].get_voted_string(), "{} out of {} ({}) eligible users voted in this poll.".format(0, 1, "0.0%"))
```

intranet/apps/polls/tests.py
```diff
@@ -57,6 +60,9 @@ def test_poll(self):
         user_choice = Answer.objects.get(question=Question.objects.get(poll=poll[0]), user=user).choice
         self.assertEqual(user_choice, Choice.objects.get(question__poll=poll[0], info="Blue"))
 
+        # Everyone can vote, one person has
+        self.assertEqual(poll[0].get_voted_string(), "{} out of {} ({}) of eligible users voted in this poll.".format(1, 1, "100.0%"))
```
theo-o, line 64:
```suggestion
        self.assertEqual(poll[0].get_voted_string(), "{} out of {} ({}) eligible users voted in this poll.".format(1, 1, "100.0%"))
```

intranet/apps/polls/tests.py
```diff
@@ -78,6 +84,9 @@ def test_can_vote(self):
 
         poll = Poll.objects.create(title="Test", description="Test", start_time=ago_5, end_time=ago_5, visible=False)
 
+        # Everyone can vote, no one has
+        self.assertEqual(poll.get_voted_string(), "{} out of {} ({}) of eligible users voted in this poll.".format(0, 1, "0.0%"))
```
theo-o, line 88:
```suggestion
        # Everyone can vote, no one has
        self.assertEqual(poll.get_voted_string(), "{} out of {} ({}) eligible users voted in this poll.".format(0, 1, "0.0%"))
```

intranet/apps/polls/tests.py
```diff
@@ -112,10 +121,16 @@ def test_can_vote(self):
         poll.groups.add(temp_group)
         self.assertFalse(poll.can_vote(user))
 
+        # temp_group can vote, no one in temp_group
+        self.assertEqual(poll.get_voted_string(), "{} out of {} ({}) of eligible users voted in this poll.".format(0, 0, "0.0%"))
```
theo-o, line 125:
```suggestion
        self.assertEqual(poll.get_voted_string(), "{} out of {} ({}) eligible users voted in this poll.".format(0, 0, "0.0%"))
```

intranet/apps/polls/tests.py
```diff
@@ -112,10 +121,16 @@ def test_can_vote(self):
...
+
         # Happening now, visible, in poll group
         user.groups.add(temp_group)
         self.assertTrue(poll.can_vote(user))
 
+        # temp_group can vote, one in temp_group
+        self.assertEqual(poll.get_voted_string(), "{} out of {} ({}) of eligible users voted in this poll.".format(0, 1, "0.0%"))
```
theo-o, line 132:
```suggestion
        self.assertEqual(poll.get_voted_string(), "{} out of {} ({}) eligible users voted in this poll.".format(0, 1, "0.0%"))
```

PR #999, opened 2020-05-15
feat(auth): ratelimit login attempts

### LINKED ISSUE #773: Ratelimit incorrect user login attempts
## Feature description
Users will be blocked from authenticating to Ion after repeated failed login attempts. This number should be fairly large to prevent honest failed attempts from failing. We would probably want to use a different limit from FCPS WiFi IPs and track by username/account.
This should also apply to the reset_password view.

Repeated login attempts (10+) are an indicator that something is up and should require manual intervention from an Ion admin or a decent cooldown period.

- See https://github.com/jazzband/django-axes for a useful library to use.

Laur04:
closes #773

## Proposed changes
- lock account after 10 failed tries for a period of 2 hours
- send emails to intranet@tjhsst.edu to notify of an account being locked
- all members of admin_all to unlock accounts

## Question/Concerns
- The original PR talks about different settings for logging in over FCPS wifi. There is no easy way to do this; it might be possible if I try to write custom methods that override the defaults, but I'm concerned about messing up the entire auth system because I think it's hard to fully test all auth situations in dev. 
- Whitelisting logins with the master password was also mentioned. The axes module is easy to whitelist usernames and ips with, but I don't really know how the admin account login works. If we want to do this, it might be easiest to whitelist ips from the vpn.
-What is the "error" color in Ion? I couldn't find it (the red I'm using now is very ugly).


requirements.txt
```diff
@@ -12,6 +12,7 @@ contextlib2==0.6.0
 cryptography==2.8
 decorator==4.4.2
 Django==2.2.11
+django-axes==5.3.1
```
theo-o, line 15: Since you are adding a new package, you should add the package & appropriate info to the alphabetical list at `docs/developing/requirements.rst`.

In addition, this build will fail, because you did not run `python setup.py egg_info`

intranet/templates/auth/emails/reset_login.html
```diff
@@ -0,0 +1,5 @@
+<p>The user <b>{{ user }}</b> has been locked out of their Ion account due to 10 failed login attempts.</p>
+<br>
```
theo-o, line 2: Why not just separate paragraphs, instead of using `<br>`

intranet/apps/auth/urls.py
```diff
@@ -5,6 +5,8 @@
 urlpatterns = [
     re_path(r"^$", views.index_view, name="index"),
     re_path(r"^login$", views.LoginView.as_view(), name="login"),
+    re_path(r"^lockout", views.lockout_view, name="lockout"),
```
theo-o, line 8: I would recommend that `lockout` & `unlock`, go at `/ratelimit`

intranet/apps/auth/views.py
```diff
@@ -103,6 +107,42 @@ def get_week_sports_school_events() -> Tuple[Container[Event], Container[Event]]
...
+    if request.method == "POST":
+        form = ResetLoginAttemptsEmailForm(request.POST)
+        if form.is_valid():
+            user = form.cleaned_data["user_account"]
+            subject = "Account Unlock Requested for {}".format(user)
+            emails = [settings.FEEDBACK_EMAIL]
+            base_url = "https://ion.tjhsst.edu/unlock"
+            data = {"user": user, "url": base_url}
+            email_send_task.delay("auth/emails/reset_login.txt", "auth/emails/reset_login.html", data, subject, emails)
+            return HttpResponseRedirect("/")
```
theo-o, line 120: Use `redirect` instead

intranet/apps/auth/views.py
```diff
@@ -103,6 +107,42 @@ def get_week_sports_school_events() -> Tuple[Container[Event], Container[Event]]
...
+    if not is_admin_all:
+        return render(request, "error/403.html", {"reason": "You are not authorized to view this page."}, status=403)
+    if request.method == "POST":
+        form = ResetLoginForm(request.POST)
+        if form.is_valid():
+            if form.cleaned_data["reset_all"]:
+                reset()
+            else:
+                reset(username=form.cleaned_data["user_account"])
+            return HttpResponseRedirect("/unlock")
```
theo-o, line 139: Use `redirect` with `reverse` instead.

intranet/apps/auth/views.py
```diff
@@ -103,6 +107,42 @@ def get_week_sports_school_events() -> Tuple[Container[Event], Container[Event]]
...
+        return render(request, "auth/unlock.html", context={"form": form, "error": True})
```
theo-o, line 140: You don't need to reimplement error handling. Django natively supports this with form validators (see [docs on `Field.error_messages`](https://docs.djangoproject.com/en/3.0/ref/forms/fields/#django.forms.Field.error_messages))

theo-o:
>     * The original PR talks about different settings for logging in over FCPS wifi. There is no easy way to do this; it might be possible if I try to write custom methods that override the defaults, but I'm concerned about messing up the entire auth system because I think it's hard to fully test all auth situations in dev.

You will need to override some methods in `django-axes` to implement this. If you conduct minimal overriding, the chance of breakage is much lower.

>     * Whitelisting logins with the master password was also mentioned. The axes module is easy to whitelist usernames and ips with, but I don't really know how the admin account login works. If we want to do this, it might be easiest to whitelist ips from the vpn.

We use a [`MasterPasswordAuthenticationBackend`](https://github.com/tjcsl/ion/blob/257c4d6b2e82d8501778ae8bcce3bc5b33f853f9/intranet/apps/auth/backends.py#L184) (see [docs on authentication backends](https://docs.djangoproject.com/en/3.0/topics/auth/customizing/#authentication-backends)) (see [where we define our authentication backends](https://github.com/tjcsl/ion/blob/257c4d6b2e82d8501778ae8bcce3bc5b33f853f9/intranet/settings/__init__.py#L314))

You will likely need to dig into `django-axes` codebase & override a few methods.

>       -What is the "error" color in Ion? I couldn't find it (the red I'm using now is very ugly).
We don't have any official one. Our "warnings" are mainly very simple pop-up boxes or random text. This is of course a non-ideal situation

intranet/apps/auth/urls.py
Laur04: So `re_path(r"^ratelimit/lockout", views.lockout_view, name="lockout"),`?

theo-o: Yes, (and same for `unlock`). You also need a trailing `$` to indicate the close of the path (since we are using `re_path` here).

PR #1008, opened 2020-06-16
fix: provision redis with docker instead of source

abagali1:
## Proposed changes
- Use docker vagrant provisioner to setup redis layer instead of building from source

Building redis-server from source threw an error when building for me and docker is more reliable and faster than building from source.

### DIFF
--- a/config/provision_vagrant.sh
+++ b/config/provision_vagrant.sh
@@ -81,14 +81,8 @@ sed -Ei "s/(^local +all +all +)peer$/\1md5/g" /etc/postgresql/10/main/pg_hba.con
 service postgresql restart
 
 # Redis
-wget --progress dot:giga http://download.redis.io/redis-stable.tar.gz
-tar xvzf redis-stable.tar.gz
-cd redis-stable
-make install
-cd utils
-echo | ./install_server.sh
-cd ../..
-rm -rf redis-stable redis-stable.tar.gz
+apt-get -y install redis-server
+systemctl enable --now redis-server
 
 # RabbitMQ
 apt-get -y install rabbitmq-server


Vagrantfile
```diff
@@ -85,6 +85,9 @@ Vagrant.configure("2") do |config|
     destination: ".ssh/#{devconfig['ssh_key']}.pub"
 
   config.vm.provision "shell", path: "config/provision_vagrant.sh"
+  config.vm.provision "docker" do |d|
+    d.run "redis", image: "redis:latest", args: "-p 6379:6379", daemonize: true
```
anonymoose2, line 89: We don't need Docker to provision Redis. I'd much rather just install the `redis-server` package.

config/provision_vagrant.sh
```diff
@@ -81,14 +81,9 @@ sed -Ei "s/(^local +all +all +)peer$/\1md5/g" /etc/postgresql/10/main/pg_hba.con
...
-tar xvzf redis-stable.tar.gz
-cd redis-stable
-make install
-cd utils
-echo | ./install_server.sh
-cd ../..
-rm -rf redis-stable redis-stable.tar.gz
+apt-get -y install redis-server
+service redis-server enable
+service redis-service start
```
anonymoose2, line 86: I'd rather use `systemctl enable --now redis-server` here. The use of `service` elsewhere is most likely legacy code from old versions of Ubuntu.

theo-o:
@anonymoose2 @abagali1 Doesn't this downgrade Redis in dev environments to Redis 5.4? This may cause inconsistencies due to differing package versions in production (6.0.4) and development.

anonymoose2:
> @anonymoose2 @abagali1 Doesn't this downgrade Redis in dev environments to Redis 4.0? This may cause inconsistencies due to differing package versions in production (6.0.4) and development.

Hmm. I don't think we use any of the Redis 6.0 features.

anonymoose2:
In addition, we should really upgrade to 20.04 soon, and it has Redis 5.0.7.

abagali1:
If it becomes a problem we can switch back to the docker provisioner and pull the `redis:6.0.4` image directly.

theo-o:
> > @anonymoose2 @abagali1 Doesn't this downgrade Redis in dev environments to Redis 4.0? This may cause inconsistencies due to differing package versions in production (6.0.4) and development.
> 
> Hmm. I don't think we use any of the Redis 6.0 features.

Even if we aren't using Redis 6.0 features, we should at least add a note in this script about the version difference, since future developers may encounter problems.

abagali1:
> > > @anonymoose2 @abagali1 Doesn't this downgrade Redis in dev environments to Redis 4.0? This may cause inconsistencies due to differing package versions in production (6.0.4) and development.
> > 
> > 
> > Hmm. I don't think we use any of the Redis 6.0 features.
> 
> Even if we aren't using Redis 6.0 features, we should at least add a note in this script about the version difference, since future developers may encounter problems.

In that case why don't we just use docker?

anonymoose2:
> > > > @anonymoose2 @abagali1 Doesn't this downgrade Redis in dev environments to Redis 4.0? This may cause inconsistencies due to differing package versions in production (6.0.4) and development.
> > > 
> > > 
> > > Hmm. I don't think we use any of the Redis 6.0 features.
> > 
> > 
> > Even if we aren't using Redis 6.0 features, we should at least add a note in this script about the version difference, since future developers may encounter problems.
> 
> In that case why don't we just use docker?

If we're going to use Docker in dev environments, then we should do a *complete* switch to Docker to match production as closely as possible.

Also, running different versions of Redis is probably the least of our worries. Dev and production are also running different versions of Postgres, RabbitMQ, and Python.

theo-o:
> If we're going to use Docker in dev environments, then we should do a _complete_ switch to Docker to match production as closely as possible.
Agreed.

PR #989, opened 2020-04-30
Eliminate settings.SENIOR_GRADUATION_YEAR

anonymoose2:
## Proposed changes
- Eliminate `settings.SENIOR_GRADUATION_YEAR` in favor of auto-detection based on current date.

## Brief description of rationale
Updating `settings.SENIOR_GRADUATION_YEAR` is annoying.

PR #1014, opened 2020-08-01
fix(dashboard): fix graduation date added in template

anonymoose2:
## Proposed changes
- Fix graduation date added in template

This breaks the graduation countdown.

In #989, I preserved the previous value of the `senior_graduation` variable (a string representation of the graduation date) but for unknown reasons I also added the `date` filter here. The `date` filter tries to operate on this string, fails, and yields an empty string. The JS tries to parse this, fails, and displays bad information to the user.


anonymoose2:
The `linting`  failure here is due to the fact that #1013 had not been merged when this PR was opened. Merging.

PR #1019, opened 2020-08-13
feat(dataimport): add delete_users management command

etnguyen03:
## Proposed changes
- Add the `delete_users` management command

To allow us to mass remove users from Ion with their FCPS student IDs, for instance, when they have withdrawn.

I will squash all commits before merging.

### DIFF
--- a/intranet/apps/dataimport/management/commands/delete_users.py
+++ b/intranet/apps/dataimport/management/commands/delete_users.py
@@ -1,0 +1,84 @@
+import csv
+import sys
+
+from django.contrib.auth import get_user_model
+from django.core.management.base import BaseCommand
+
+
+class Command(BaseCommand):
+    help = "Given a list of FCPS student IDs, delete them if they exist."
+
+    def add_arguments(self, parser):
+        parser.add_argument("--filename", dest="filename", type=str, help="Filename to import data from. Must supply either this or 'student-id'.")
+        parser.add_argument(
+            "--column-header",
+            dest="header",
+            default="Student ID",
+            type=str,
+            nargs=1,
+            help="Header associated with the student ID column. Use with the 'filename' option",
+        )
+        parser.add_argument(
+            "--student-id", dest="student_ids", nargs="+", help="A space-separated list of student IDs to remove. Must supply this or 'filename'."
+        )
+        parser.add_argument("--run", action="store_true", dest="run", default=False, help="Actually run.")
+        parser.add_argument("--confirm", action="store_true", dest="confirm", default=False, help="Skip confirmation.")
+
+    def ask(self, q):
+        if input("{} [Yy]: ".format(q)).lower() != "y":
+            self.stdout.write(self.style.ERROR("Abort."))
+            sys.exit()
+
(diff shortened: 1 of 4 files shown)


intranet/apps/dataimport/management/commands/delete_users.py
```diff
@@ -0,0 +1,88 @@
...
+        # Read in the list of student IDs, either from args or from the given filename
+        if options["student_ids"] is None:
+            ids = self.read_student_ids(options["filename"], options["header"])
+        else:
+            ids = options["student_ids"]
+
+        # Get a list of user model instances
+        users = []
+        for student_id in ids:
+            user = get_user_model().objects.filter(student_id=student_id).first()
```
Laur04, line 64: Why are you using `filter()` here and not `get()`? `student_id` is unique, so I think `get()` is more appropriate.

intranet/apps/dataimport/management/commands/delete_users.py
```diff
@@ -0,0 +1,88 @@
...
+            if user is not None:
+                users.append(user)
```
Laur04, line 66: You could replace this with:
```
if get_user_model().objects.get(student_id=student_id).exist():
  users.append(get_user_model().objects.get(student_id=student_id))
```

intranet/apps/dataimport/management/commands/delete_users.py
etnguyen03: My reasoning was that `get()` would raise a `DoesNotExist` if the given user(s) do not exist, which isn't the case with the data we're working with (some given users were already deleted).

From what I've seen elsewhere, the two ways to deal with this problem are a try-catch or the `filter()` that I used.

abagali1: You can also do .filter(student_id__in=ids) which will just return a QuerySet of all the students who have a student_id in ids

abagali1: You can also do .filter(student_id__in=ids) which will just return a QuerySet of all the students who have a student_id in ids

Laur04: Yea that’s probably the best way to minimize queries.

etnguyen03:
@Laur04?

intranet/apps/dataimport/management/commands/delete_users.py
```diff
@@ -0,0 +1,84 @@
...
+        if options["student_ids"] is None:
+            ids = self.read_student_ids(options["filename"], options["header"])
+        else:
+            ids = options["student_ids"]
+
+        # Get a list of user model instances
+        users = get_user_model().objects.filter(student_id__in=ids)
+
+        self.stdout.write(self.style.WARNING("These users are about to be deleted."))
+        self.stdout.write(str([u.username for u in users]))
```
theo-o, line 65:
```suggestion
        self.stdout.write(str([users.values_list("username", flat=True)]))
```
Minor performance nitpick

PR #1020, opened 2020-08-14
feat: declutter nav dropdown menu

### LINKED ISSUE #904: Declutter navigation dropdown menu
## Feature description
On the top navigation dropdown (located at the top right of pages), we have a listing of various links. We should  move some of the links to a generic preferences heading that is collapsible.

The dropdown is getting cluttered and some of the elements are all part of "user preferences".

- Move "Item Registration", "Lost & Found", and "Preferences" under "User Settings"

DasBundesland:
closes #904

## Proposed changes
- Move the "Preferences", "Item Registration", "Lost and Found", and "OAuth" links in the nav dropdown menu under a "User Settings" header that is collapsible.

- To make the dropdown menu more organized and less full (see #904)

### DIFF
--- a/intranet/templates/page_with_header.html
+++ b/intranet/templates/page_with_header.html
@@ -92,15 +92,25 @@
                     </div>
                     <ul class="dropdown-menu">
                         <div class="arrow"></div>
-                        <li><a href="{% url 'user_profile'%}"><i class="fas fa-user"></i> My Profile</a></li>
-                        <li><a href="{% url 'preferences' %}"><i class="fas fa-cog"></i> Preferences</a></li>
-                        {% if not request.user.is_restricted %}
-                        <li><a href="{% url 'itemreg' %}"><i class="fas fa-edit"></i> Item Registration</a></li>
-                        <li><a href="{% url 'lostfound' %}"><i class="far fa-compass"></i> Lost and Found</a></li>
-                        <li><a href="{% url 'oauth2_provider:list' %}"><i class="fas fa-key"></i> OAuth</a></li>
-                        {% endif %}
+                        <li>
+                            <div style="float:left;">
+                                <a href="{% url 'user_profile'%}"><i class="fas fa-user"></i> My Profile</a>
+                            </div>
+                            <div class="has-dropdown-subgroup">
+                                <a class="subgroup-arrow-parent" href="#"><i class="fas fa-caret-down subgroup-arrow"></i></a>
+                                <ul class="dropdown-subgroup" style="display:none;">
+                                    <li><a href="{% url 'preferences' %}"><i class="fas fa-cog"></i> Preferences</a></li>
+                                    {% if not request.user.is_restricted %}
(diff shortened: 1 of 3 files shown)


Ion.egg-info/SOURCES.txt
```diff
@@ -689,6 +689,7 @@ intranet/apps/users/migrations/0031_auto_20190715_1259.py
 intranet/apps/users/migrations/0032_userdarkmodeproperties.py
 intranet/apps/users/migrations/0033_user_last_global_logout_time.py
 intranet/apps/users/migrations/0034_remove_userdarkmodeproperties__dark_mode_unlocked.py
+intranet/apps/users/migrations/0035_auto_20200810_1547.py
```
Laur04, line 692: Have you rebased against the most recent dev?

intranet/templates/page_with_header.html
```diff
@@ -112,5 +118,6 @@ <h1>Intranet</h1>
     <div class="main">
         {% block content %}
         {% endblock %}
+
```
Laur04, line 121: Remove extra whitespace

Laur04:
`Session Management` should fall under the dropdown as well, since that is also a "user preference." 

I think the caret looks a little strange because everything else has an icon. I would suggest moving everything that is currently under `User Settings` under a dropdown from `My Profile` with the carot on the right. Users would still be able to click on `My Profile` and be taken to their profile, but they would also have the option to arrow down and click a dropdown option. What are your thoughts?

DasBundesland:
I think that's a good idea. I also thought the caret looked out-of-place, but I wasn't sure how to make it cleaner. I'll work on those changes. Thanks!

Laur04:
LGTM. @etnguyen03 @abagali1?

intranet/templates/page_with_header.html
```diff
@@ -92,15 +92,25 @@ <h1>Intranet</h1>
...
-                        <li><a href="{% url 'itemreg' %}"><i class="fas fa-edit"></i> Item Registration</a></li>
-                        <li><a href="{% url 'lostfound' %}"><i class="far fa-compass"></i> Lost and Found</a></li>
-                        <li><a href="{% url 'oauth2_provider:list' %}"><i class="fas fa-key"></i> OAuth</a></li>
-                        {% endif %}
+                        <li>
+                            <div style="float:left;">
+                                <a href="{% url 'user_profile'%}"><i class="fas fa-user"></i> My Profile</a>
+                            </div>
+                            <div class="has-dropdown-subgroup">
+                                <a class="subgroup-arrow-parent"><i class="fas fa-caret-down subgroup-arrow"></i></a>
```
etnguyen03, line 100:
```suggestion
                                <a class="subgroup-arrow-parent" href="#"><i class="fas fa-caret-down subgroup-arrow"></i></a>
```

Kind of a nitpick, but if one were to hover over the arrow without this, the cursor would not turn into a "thumb" (or otherwise show as clickable).

PR #1024, opened 2020-08-15
feat(dataimport): add import_students and import_staff command

etnguyen03:
## Proposed changes
- Add commands to import students (`import_students`) and staff (`import_staff`)

We don't have a command to bulk import students and staff from CSVs, or at least as far as I'm aware. And we need one soon for new staff/students.

### DIFF
--- a/intranet/apps/dataimport/management/commands/import_staff.py
+++ b/intranet/apps/dataimport/management/commands/import_staff.py
@@ -1,0 +1,63 @@
+import csv
+import sys
+
+from django.contrib.auth import get_user_model
+from django.core.management.base import BaseCommand
+
+from ....users.models import Email
+
+
+class Command(BaseCommand):
+    help = (
+        "Given an CSV of staff, add them to Ion. Required columns include 'Username', 'First Name', 'Last Name', "
+        "and 'Middle Name'. Optional columns include 'Nick Name' and 'Gender'."
+    )
+
+    def add_arguments(self, parser):
+        parser.add_argument("--filename", dest="filename", type=str, required=True, help="Filename to import data from. Required.")
+        parser.add_argument("--run", action="store_true", dest="run", default=False, help="Actually run.")
+        parser.add_argument("--confirm", action="store_true", dest="confirm", default=False, help="Skip confirmation.")
+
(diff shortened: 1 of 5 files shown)


intranet/apps/dataimport/management/commands/import_staff.py
```diff
@@ -0,0 +1,63 @@
...
+        if do_run:
+            if not options["confirm"]:
+                self.ask("Add new users?")
+        else:
+            self.stdout.write(self.style.WARNING("In pretend mode."))
+
+        # Loop through our new users
+        for new_user in data:
+            nickname = new_user["Nick Name"].strip() if "Nick Name" in new_user.keys() else ""
+            gender = new_user["Gender"].strip() == "M" if "Gender" in new_user.keys() else None  # TODO: is it "M" or "Male" or something else?
```
Laur04, line 41: I don't know that we record gender currently...most users don't seem to have it.

intranet/apps/dataimport/management/commands/import_students.py
```diff
@@ -0,0 +1,145 @@
...
+                # Check to see if username already taken
+                if new_user["TJHSST_username"] in existing_usernames:
+                    new_user["TJHSST_username"] = self.find_next_available_username(new_user["TJHSST_username"], existing_usernames)
+
+                # Now we can safely add this user
+                if do_run:
+                    counselor = get_user_model().objects.get(username=new_user["Counselor"].strip()) if "Counselor" in new_user.keys() else None
+                    nickname = new_user["Nick Name"].strip() if "Nick Name" in new_user.keys() else ""
+                    gender = (
+                        new_user["Gender"].strip() == "M" if "Gender" in new_user.keys() else None
```
Laur04, line 122: As above

intranet/apps/dataimport/management/commands/import_staff.py
etnguyen03: True, but it doesn't hurt - either way, if we don't supply a gender column, it would be entered as `None`.

The two users I'm importing I'm told were female (the email mentions them as "Ms. {{ NAME }}") so we may as well put that in while we're at it

PR #1041, opened 2020-09-10
feat(emailforward): add feedback for forward button

### LINKED ISSUE #1040: Feedback when pressing button for forward email
## Feature description
Currently, there is no feedback for the "forward email" button. As it is pressed, there is no message that says "processing."

Many people are spamming the button until it works or fails, which is bad.

- Possibly handle it similarly to Ion's login button? (where the button swaps with a spinner)
- Disable the button while Ion is waiting a response

Laur04:
closes #1040

Of the three options, I went with imitating the login spinner because you can't disable `<a></a>` tags and @anonymoose2's suggestion (as I read it) doesn't actually prevent users from re-clicking the button.


Laur04:
Codacy is being picky, but I don't think it raises anything valid; everything else passes.

etnguyen03 (approved):
LGTM

Codacy failing is fine with me; formatting on the CSS is consistent with the rest of the file and `Spinner` is defined elsewhere.

intranet/templates/preferences/preferences.html
```diff
@@ -69,6 +69,7 @@
         {% stylesheet 'dark/nav' %}
         {% stylesheet 'dark/preferences' %}
     {% endif %}
+    <script src="{% static 'js/vendor/spin.min.js' %}"></script>
```
theo-o, line 72: This should go in the `js` block above

intranet/templates/preferences/preferences.html
```diff
@@ -108,6 +109,7 @@ <h2>Mail Forwarding</h2>
             <tr>
               <td>
                 <a id="mail-forwarding-submit" href="#" class="button">Forward Email</a>
+                <div class='spinner-container'></div>
```
theo-o, line 112:
```suggestion
                <div class="spinner-container"></div>
```

Extremely minor nit

PR #1049, opened 2020-10-08
feat(eighth): redirect to info page near eighth period

Laur04:
## Proposed changes
- redirect to `/eighth/activity/location` starting 20 minutes before the first 8th block of a day and ending 20 minutes after that block's start time.
- `/eighth/activity/location` details the two upcoming 8th blocks, including title, location, sponsors, comments and descriptions

- avoid loading `/` or `/eighth` for students checking 8th period codes during peak times

### DIFF
--- a/intranet/apps/eighth/views/signup.py
+++ b/intranet/apps/eighth/views/signup.py
@@ -6,6 +6,7 @@ from prometheus_client import Summary
 
 from django import http
 from django.conf import settings
+from django.contrib import messages
 from django.contrib.auth import get_user_model
 from django.contrib.auth.decorators import login_required
 from django.db import transaction
@@ -363,6 +364,24 @@ def toggle_favorite_view(request):
         else:
             activity.favorites.add(request.user)
             return http.HttpResponse("Favorited activity.")
+
+
+@login_required
+def eighth_location(request):
+    blocks = EighthBlock.objects.get_blocks_today()
+    if blocks:
+        sch_acts = []
+        for b in blocks:
+            try:
+                act = request.user.eighthscheduledactivity_set.get(block=b)
+                sch_acts.append([b, act, ", ".join([r.name for r in act.get_true_rooms()]), ", ".join([s.name for s in act.get_true_sponsors()])])
+            except EighthScheduledActivity.DoesNotExist:
+                sch_acts.append([b, None])
+
+        return render(request, "eighth/location.html", context={"sch_acts": sch_acts})
+    else:
+        messages.error(request, "There are no eighth period blocks scheduled today.")
+        return redirect("index")
 
 
 @require_POST
--- a/intranet/settings/__init__.py
+++ b/intranet/settings/__init__.py
@@ -42,7 +42,8 @@ NOMINATION_POSITION = ""
 ENABLE_WAITLIST = False  # WARNING: Enabling the waitlist causes severe performance issues
 ENABLE_BUS_APP = True
 ENABLE_BUS_DRIVER = True
-ENABLE_PRE_EIGHTH_REDIRECT = False
+ENABLE_PRE_EIGHTH_CLOSE_SIGNUP_REDIRECT = False
+ENABLE_PRE_EIGHTH_LOCATION_REDIRECT = True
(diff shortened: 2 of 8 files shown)


intranet/settings/__init__.py
```diff
@@ -42,7 +42,8 @@
 ENABLE_WAITLIST = False  # WARNING: Enabling the waitlist causes severe performance issues
 ENABLE_BUS_APP = True
 ENABLE_BUS_DRIVER = True
-ENABLE_PRE_EIGHTH_REDIRECT = False
+ENABLE_PRE_EIGHTH_SIGNUP_REDIRECT = False
```
anonymoose2, line 45: If we're in the business of renaming this option, I would propose instead `ENABLE_PRE_EIGHTH_CLOSE_SIGNUP_REDIRECT`.

Laur04: Done

intranet/templates/eighth/location.html
```diff
@@ -0,0 +1,53 @@
...
+        {% stylesheet 'dark/base' %}
+        {% stylesheet 'dark/nav' %}
+    {% endif %}
+{% endblock %}
+
+{% block main %}
+<div class="primary-content" id="quick-schedule">
+    <h2>Your upcoming eighth periods are:</h2>
+    <br>
+    <h3>{{ blocks.0.block_letter }} Block ({{ blocks.0.date }})</h3>
```
theo-o, line 29: To be consistent with the rest of Ion, we should hyperlink either this block information or the activity name below to the eighth signup page for this block.

intranet/templates/eighth/location.html
```diff
@@ -0,0 +1,53 @@
...
+    {% if dark_mode_enabled %}
+        {% stylesheet 'dark/base' %}
+        {% stylesheet 'dark/nav' %}
+    {% endif %}
+{% endblock %}
+
+{% block main %}
+<div class="primary-content" id="quick-schedule">
+    <h2>Your upcoming eighth periods are:</h2>
+    <br>
```
theo-o, line 28: If you apply my change mentioned in `eighth/views/activities.py`, the below can be done via a for loop

intranet/apps/eighth/views/activities.py
```diff
@@ -47,6 +47,28 @@ def activity_view(request, activity_id=None):
     return render(request, "eighth/activity.html", context)
 
 
+@login_required
+def eighth_location(request):
```
theo-o, line 51: I'm not 100% sure if `activities.py` is the right place for this view. Perhaps `signup.py` would be better?

intranet/apps/eighth/views/activities.py
```diff
@@ -47,6 +47,28 @@ def activity_view(request, activity_id=None):
...
+    blocks = EighthBlock.objects.get_upcoming_blocks(2)
+    if len(blocks) >= 2:
```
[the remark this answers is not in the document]
Laur04: I actually think `get_blocks_today` might work better. The view shouldn't be called if there aren't 8th blocks that day, but I'll put checks to makes sure that non-red days don't cause issues.

intranet/apps/eighth/views/signup.py
```diff
@@ -365,6 +365,23 @@ def toggle_favorite_view(request):
...
+        for b in blocks:
+            try:
+                act = request.user.eighthscheduledactivity_set.get(block=b)
+                sch_acts.append([b, act, ", ".join([r.name for r in act.get_true_rooms()]), ", ".join([s.name for s in act.get_true_sponsors()])])
+            except EighthScheduledActivity.DoesNotExist:
+                sch_acts.append([b, None])
+
+        return render(request, "eighth/location.html", context={"sch_acts": sch_acts})
+    else:
+        return redirect("index")
```
etnguyen03, line 382:
```suggestion
        messages.error(request, "There are no eighth period blocks scheduled today.")
        return redirect("index")
```

etnguyen03 (approved):
LGTM - maybe we may want to add the line below so if someone stumbles across the `/eighth/glance` page on a non-Eighth day (or something similar to it)

PR #1050, opened 2020-10-14
fix(eighth): change times for eighth_location to be shown

Laur04:
Basically changes the show time for `eighth_location` from within 20 minutes of the first block start time to 30 minutes before first block start through 20 minutes after last block start.

Forgot to PR this last week, sorry.

### DIFF
--- a/intranet/apps/auth/views.py
+++ b/intranet/apps/auth/views.py
@@ -200,11 +200,13 @@ class LoginView(View):
                     today_8 = Day.objects.today().day_type.blocks.filter(name__contains="8")
                     if today_8:
                         first_start_time = time(today_8[0].start.hour, today_8[0].start.minute)
+                        last_start_time = time(today_8.last().start.hour, today_8.last().start.minute)
                         first_start_date = datetime.combine(now.today(), first_start_time)
+                        last_start_date = datetime.combine(now.today(), last_start_time)
                         if (
-                            first_start_date - timedelta(minutes=20)
+                            first_start_date - timedelta(minutes=30)
                             < datetime.combine(now.today(), now.time())
-                            < first_start_date + timedelta(minutes=20)
+                            < last_start_date + timedelta(minutes=20)
                         ):
                             default_next_page = "eighth_location"
                 except AttributeError:


intranet/apps/auth/views.py
```diff
@@ -200,11 +200,13 @@ def post(self, request):
                     today_8 = Day.objects.today().day_type.blocks.filter(name__contains="8")
                     if today_8:
                         first_start_time = time(today_8[0].start.hour, today_8[0].start.minute)
+                        last_start_time = time(today_8.pop().start.hour, today_8.pop().start.minute)
```
etnguyen03, line 203:
```suggestion
                        last_start_time = time(today_8[-1].start.hour, today_8[-1].start.minute)
```

`.pop()` would remove the element before returning it; causing the second `.pop()` to not return the same block as the first `.pop()`.

etnguyen03 (changes requested):

intranet/apps/auth/views.py
```diff
@@ -200,11 +200,14 @@ def post(self, request):
                     today_8 = Day.objects.today().day_type.blocks.filter(name__contains="8")
                     if today_8:
                         first_start_time = time(today_8[0].start.hour, today_8[0].start.minute)
+                        last_block = today_8.pop()
```
etnguyen03, line 203: 'QuerySet' object has no attribute 'pop'

```suggestion
                        last_block = today_8.last()
```

Perhaps you may want to revert and go back to `today_8.last().start.hour` and the same for the minute?

etnguyen03 (changes requested):

intranet/apps/auth/views.py
Laur04: Trying to figure out why my dev env didn't catch this..ugh

etnguyen03 (approved):

intranet/apps/auth/views.py
etnguyen03: I have a feeling it is because of the try-catch used; I had to do a `print` to get the exception that was thrown

PR #1052, opened 2020-10-15
fix(eighth): fix duplicate signups when transferring students

### LINKED ISSUE #1051: Moving an activity causes duplicate signups
## Steps to reproduce
1. Two activities on two separate blocks
2. Have a student sign up for the two activities
3. Move one of the activities to the other block

## Expected behavior
Either one of the signups is dropped, or something is shown so the admin can choose which signup is kept.

## Actual behavior
A duplicate signup is created.

Laur04:
closes #1051

### DIFF
--- a/intranet/apps/eighth/tasks.py
+++ b/intranet/apps/eighth/tasks.py
@@ -67,6 +67,52 @@ def room_changed_single_email(
 
 
 @shared_task
+def transferred_activity_email(
+    dest_act: EighthScheduledActivity, source_act: EighthScheduledActivity, duplicate_students  # pylint: disable=unsubscriptable-object
+):  # pylint: disable=unsubscriptable-object
+    """Notifies all the users already signed up for an EighthScheduledActivity that they have been transferred into a new activity.
+
+    Args:
+        dest_act: The activity that the students were transferred into.
+        dest_act: The activity that the students were transferred from.
+        duplicate_students: The students that were transferred into a new activity.
+
+    """
+    date_str = dest_act.block.date.strftime("%A, %B %-d")
+
+    # Emails are sent regardless of notification settings so that students have the chance to ensure that they are signed up for the correct place.
+    emails = [u.notification_email for u in duplicate_students]
+
+    if not emails:
+        return
+
+    base_url = "https://ion.tjhsst.edu"
+
+    data = {
+        "dest_act": dest_act,
+        "source_act": source_act,
+        "date_str": date_str,
+        "base_url": base_url,
+    }
+
+    logger.debug(
+        "Transferring students from %s to %s resulted in duplicate signups being deleted; emailing the %d affected users",
+        source_act,
+        dest_act.id,
+        len(emails),
+    )
+
+    email_send(
+        "eighth/emails/transferred_activity.txt",
+        "eighth/emails/transferred_activity.html",
+        data,
+        "8th Period Transfer to {} on {}".format(dest_act.activity.name, date_str),
+        emails,
+        bcc=True,
+    )
+
+
+@shared_task
 def room_changed_activity_email(
     act: EighthActivity, old_rooms: Collection[EighthRoom], new_rooms: Collection[EighthRoom]  # pylint: disable=unsubscriptable-object
 ):  # pylint: disable=unsubscriptable-object
--- a/intranet/apps/eighth/views/admin/scheduling.py
+++ b/intranet/apps/eighth/views/admin/scheduling.py
@@ -5,19 +5,21 @@ from formtools.wizard.views import SessionWizardView
 
 from django.contrib import messages
 from django.core.management import call_command
+from django.db import transaction
 from django.db.models import Count
 from django.db.models.manager import Manager
 from django.forms.formsets import formset_factory
 from django.http import Http404
 from django.shortcuts import redirect, render
 
+from .....utils.locking import lock_on
 from .....utils.serialization import safe_json
 from ....auth.decorators import eighth_admin_required
 from ...forms.admin.activities import ActivitySelectionForm
 from ...forms.admin.blocks import BlockSelectionForm
 from ...forms.admin.scheduling import ScheduledActivityForm
 from ...models import EighthActivity, EighthBlock, EighthRoom, EighthScheduledActivity, EighthSignup, EighthSponsor
-from ...tasks import room_changed_single_email
+from ...tasks import room_changed_single_email, transferred_activity_email
 from ...utils import get_start_date
 
 logger = logging.getLogger(__name__)
@@ -402,9 +404,18 @@ def transfer_students_action(request):
     else:
         raise Http404
 
+    send_emails = bool(request.POST.get("send_emails"))
     num = source_act.members.count()
 
-    context = {"admin_page_title": "Transfer Students", "source_act": source_act, "dest_act": dest_act, "dest_unsignup": dest_unsignup, "num": num}
+    context = {
+        "admin_page_title": "Transfer Students",
+        "source_act": source_act,
+        "dest_act": dest_act,
+        "dest_unsignup": dest_unsignup,
+        "num": num,
+        "moved_students": None,
+        "send_emails": send_emails,
+    }
 
     if request.method == "POST":
         if dest_unsignup and not dest_act:
@@ -412,14 +423,34 @@ def transfer_students_action(request):
             invalidate_obj(source_act)
             messages.success(request, "Successfully removed signups for {} students.".format(num))
         else:
-            source_act.eighthsignup_set.update(scheduled_activity=dest_act)
+            # In order to prevent duplicate signups when transferring students between activities in different blocks,
+            # we need to delete any `EighthSignup`s already present in the block of `dest_act` for transferred students.
+            duplicate_sign_up_users = []
+            duplicate_signups = []
+            with transaction.atomic():
+                lock_on(source_act.members.all())
+                for u in source_act.members.all():
+                    conflict_signup = u.eighthsignup_set.filter(scheduled_activity__block=dest_act.block)
+                    if conflict_signup.exists():
+                        duplicate_sign_up_users.append(u)
+                        duplicate_signups.extend(list(conflict_signup))
+                for signup in duplicate_signups:
+                    signup.delete()
+                    logger.debug("Deleted %s signup", signup)
+
+                source_act.eighthsignup_set.update(scheduled_activity=dest_act)
 
             invalidate_obj(source_act)
             invalidate_obj(dest_act)
             messages.success(request, "Successfully transfered {} students.".format(num))
-        return redirect("eighth_admin_dashboard")
-    else:
-        return render(request, "eighth/admin/transfer_students.html", context)
+
+            if duplicate_sign_up_users:
+                if send_emails:
+                    transferred_activity_email.delay(dest_act, source_act, duplicate_sign_up_users)
+                context["moved_students"] = duplicate_signups
+            else:
+                return redirect("eighth_admin_dashboard")
+    return render(request, "eighth/admin/transfer_students.html", context)
 
 
 @eighth_admin_required
--- a/intranet/templates/eighth/admin/transfer_students.html
+++ b/intranet/templates/eighth/admin/transfer_students.html
@@ -30,6 +30,17 @@
                 {% endif %}
             {% endcomment %}
         </form>
+    {% elif moved_students %}
+        <p>
+            The following students were already signed up for activities on {{ dest_act.block }}.
+            These signups have been replaced with {{ dest_act }} and the students have
+            {% if not send_emails %}<b>not</b>{% endif %} been emailed:
+        </p>
+        <ul>
+            {% for s in moved_students %}
+            <li>{{ s }}</li>
+            {% endfor %}
+        </ul>
     {% else %}
         {% if dest_unsignup %}
             <p>Do you want to remove the signups for <b>{{ num }} students</b> from:</p>
@@ -38,9 +49,13 @@
         {% else %}
             <p>Do you want to transfer <b>{{ num }} students</b> from:</p>
             <p><b>{{ source_act }}</b></p>
-            to:<br><br>
+            to:<br>
             <p><b>{{ dest_act }}</b></p>
             <br>
+            <p>
+                <b>WARNING:</b> If the students are already signed up for an activity during {{ dest_act.block }},
+                the previous signups will be deleted. Notify students of this change by enabling "Notify Students" below.
+            </p>
         {% endif %}
         <form action="{% url 'eighth_admin_transfer_students_action' %}" method="post">
             {% csrf_token %}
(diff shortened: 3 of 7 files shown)


intranet/apps/eighth/views/admin/scheduling.py
```diff
@@ -404,22 +406,44 @@ def transfer_students_action(request):
...
         else:
-            source_act.eighthsignup_set.update(scheduled_activity=dest_act)
+            duplicate_sign_up_users = []
+            duplicate_signups = []
+            with transaction.atomic():
+                for u in source_act.members.all():
+                    conflict_signup = u.eighthsignup_set.filter(scheduled_activity__block=dest_act.block)
+                    if conflict_signup.exists():
+                        duplicate_sign_up_users.append(u)
+                        duplicate_signups.append(conflict_signup[0])
```
theo-o, line 431: While we hope that there are no duplicate signup bugs elsewhere in Ion, assuming that fact might not be a great idea.

theo-o (commented):
This change may be worthy of unit tests to prevent future regressions and ensure consistent behavior

intranet/apps/eighth/tasks.py
```diff
@@ -66,6 +66,49 @@ def room_changed_single_email(
     )
 
 
+@shared_task
+def transferred_activity_email(
+    dest_act: EighthScheduledActivity, duplicate_students: Collection[get_user_model()]  # pylint: disable=unsubscriptable-object
```
anonymoose2, line 71:
```suggestion
    dest_act: EighthScheduledActivity, duplicate_students  # pylint: disable=unsubscriptable-object
```
Type annotations generally aren't supposed to contain function calls.

In the past, I've just omitted type annotations on parameters that reference `User` in `eighth/models.py` 
-- `eighth/models.py` depends on `users/models.py`, so importing it would create circular import issues. I've since learned that it's possible to work around the circular import issue, but that would probably be a matter for another PR.

intranet/apps/eighth/views/admin/scheduling.py
```diff
@@ -404,22 +406,44 @@ def transfer_students_action(request):
...
+                source_act.eighthsignup_set.update(scheduled_activity=dest_act)
+
+            transferred_activity_email.delay(dest_act, duplicate_sign_up_users)
+
+            context["moved_students"] = duplicate_signups
 
             invalidate_obj(source_act)
             invalidate_obj(dest_act)
             messages.success(request, "Successfully transfered {} students.".format(num))
-        return redirect("eighth_admin_dashboard")
```
anonymoose2, line 420: I would keep the redirect if there are no duplicates, but render a different "transfer_students_results" page if there are any duplicates.

intranet/apps/eighth/views/admin/scheduling.py
```diff
@@ -404,22 +406,44 @@ def transfer_students_action(request):
...
-            source_act.eighthsignup_set.update(scheduled_activity=dest_act)
+            duplicate_sign_up_users = []
+            duplicate_signups = []
+            with transaction.atomic():
+                for u in source_act.members.all():
+                    conflict_signup = u.eighthsignup_set.filter(scheduled_activity__block=dest_act.block)
+                    if conflict_signup.exists():
+                        duplicate_sign_up_users.append(u)
+                        duplicate_signups.append(conflict_signup[0])
+                lock_on(duplicate_sign_up_users)
```
anonymoose2, line 432: I would lock on **all** of the `User`s that are being transferred, to ensure that nothing happens concurrently.

anonymoose2 (changes requested):

intranet/apps/eighth/views/admin/scheduling.py
anonymoose2: @theo-o what are you suggesting? Add all the users, then check for duplicates retroactively?

theo-o: I am suggesting adding all elements of the `duplicate_signups` QuerySet, instead of just the first element.

intranet/apps/eighth/views/admin/scheduling.py
Laur04: I think it is useful to have the transfer messages appear beneath the original text and this stays in theme with many of the 8th admin functions, which all have the appearance of occurring on the same page. Redirecting to another page would also involve repeating a lot of code that is already in this view. I'd rather just redirect to "eighth_admin_dashboard" if there are no duplicates.

etnguyen03: My concern is that such a "return the same-page" may cause an admin to press the "Transfer" button again; although that won't be a concern (as pressing the "Transfer" button again doesn't appear to break anything), but I think that it'd be better off to redirect to somewhere without a "Transfer" button.

Laur04: Okay the latest rendition (just pushed) keeps it on the same page as far as django is concerned, but hides the "Transfer" section with template tags. This mimics the behavior of the rest of the page, which uses template tags to show/hide the wizard.

anonymoose2: I'm happy with that. @etnguyen03 ?

intranet/templates/eighth/emails/transferred_activity.html
```diff
@@ -0,0 +1,11 @@
+<html><body>
+    {% with description="Room changes" target=info_link name="View status" %}
+        {% include "email_metadata.html" %}
+    {% endwith %}
+
+    <p>
+        You were transferred by the 8th Period Office into {{ dest_act.activity }} on {{ date_str }} during {{ dest_act.block.block_letter }} block. Your previous signup has been removed, but you may re-signup on Ion if you wish.
```
anonymoose2, line 7:
```suggestion
        You were transferred by the 8th Period Office into {{ dest_act.activity }} on {{ date_str }} during {{ dest_act.block.block_letter }} block. Your previous signup during this block has been removed, but you may re-signup on Ion if you wish.
```
Just a minor clarification.

intranet/templates/eighth/emails/transferred_activity.txt
```diff
@@ -0,0 +1,5 @@
+You were transferred by the 8th Period Office into {{ dest_act.activity }} on {{ date_str }} during {{ dest_act.block.block_letter }} block. Your previous signup has been removed, but you may re-signup on Ion if you wish.
```
anonymoose2, line 1:
```suggestion
You were transferred by the 8th Period Office into {{ dest_act.activity }} on {{ date_str }} during {{ dest_act.block.block_letter }} block. Your previous signup during this block has been removed, but you may re-signup on Ion if you wish.
```

intranet/apps/eighth/views/admin/scheduling.py
etnguyen03: Good to me as well.

intranet/templates/eighth/admin/transfer_students.html
```diff
@@ -38,9 +48,14 @@
         {% else %}
             <p>Do you want to transfer <b>{{ num }} students</b> from:</p>
             <p><b>{{ source_act }}</b></p>
-            to:<br><br>
+            to:<br>
```
anonymoose2, line 51: Why was this `<br>` removed?

intranet/templates/eighth/admin/transfer_students.html
```diff
@@ -30,6 +30,16 @@
...
             {% endcomment %}
         </form>
+    {% elif moved_students %}
+        <p>
+            The following students were already signed up for activities on {{ dest_act.block }}.
+            These signups have been replaced with {{ dest_act }}, and these students have been notified via email:
+        </p>
+        <ul>
+            {% for s in moved_students %}
+            <li>{{ s }}</li>
```
anonymoose2, line 40: I would want to print out the activity that each student was in. That way, if the admin who performs this action realizes this change messes something up, they can easily go through and revert the changes.

etnguyen03:
One thing that I found: if you have two activities (call them X and Y) on blocks A and B (respectively), then if a user is signed up for X for A block and Y for B block, and someone starts a "transfer" from Y B block to X A block, then users will be emailed as if their A block signup changed, but when in reality only their B block signup changed.

So we may want to change the email to read something like "your B block (or the block from which you were transferred from) signup was cancelled; sign up before the deadline", or maybe we should just email everyone affected when a transfer happens?

intranet/templates/eighth/admin/transfer_students.html
Laur04: This does. It prints the `EighthSignup` which includes the username, activity and block.

intranet/templates/eighth/admin/transfer_students.html
Laur04: The space was weird. You can see it without a dev env at https://ion.tjhsst.edu/eighth/admin/scheduling/transfer_students_action?source_act=135085&dest_act=133549. This just moves that bottom line up.

Laur04:
> One thing that I found: if you have two activities (call them X and Y) on blocks A and B (respectively), then if a user is signed up for X for A block and Y for B block, and someone starts a "transfer" from Y B block to X A block, then users will be emailed as if their A block signup changed, but when in reality only their B block signup changed.
> 
> So we may want to change the email to read something like "your B block (or the block from which you were transferred from) signup was cancelled; sign up before the deadline", or maybe we should just email everyone affected when a transfer happens?

We could make it even more general..."The eighth period office transferred you into a new activity on {{ block }}. This resulted in you being moved out of one or more activities that you previously signed up for. Please check your signups on Ion to ensure that they are what you want."

Since we are overriding email preferences here in order to give what we consider important information, I don't think we should email anyone except the students who had signups deleted.

etnguyen03:
> > One thing that I found: if you have two activities (call them X and Y) on blocks A and B (respectively), then if a user is signed up for X for A block and Y for B block, and someone starts a "transfer" from Y B block to X A block, then users will be emailed as if their A block signup changed, but when in reality only their B block signup changed.
> > So we may want to change the email to read something like "your B block (or the block from which you were transferred from) signup was cancelled; sign up before the deadline", or maybe we should just email everyone affected when a transfer happens?
> 
> We could make it even more general..."The eighth period office transferred you into a new activity on {{ block }}. This resulted in you being moved out of one or more activities that you previously signed up for. Please check your signups on Ion to ensure that they are what you want."
> 
> Since we are overriding email preferences here in order to give what we consider important information, I don't think we should email anyone except the students who had signups deleted.

Well if 8PO (or someone) does it after someone signs up for eighth period, they may not check Ion until after the signup deadline and forget to sign up for an activity for which they were assigned out of.

I get the override email preferences but I still think that if a student is un-signedup from an activity, they should get an email. That being said, I'm not sure if 8PO will ever use the "transfer" to transfer the entire school out of a sticky activity (like, for instance, if they want to move the A block tomorrow to B block); if they do, then we'd be emailing the entire school which probably isn't a good idea to do. Maybe have a checkbox ("do you want to email everyone affected?") before sending - perhaps on the "confirm" page?

intranet/templates/eighth/admin/transfer_students.html
anonymoose2: Ah. Missed that.

Laur04:
Okay, how about this? To avoid the issue you brought up with sticky, school-wide activities while still ensuring that people don't get stuck in something they didn't want after the deadline, I will send an email to anyone who had an `EighthSignup` deleted in the transfer process, unless the activity associated with that `EighthSignup` is sticky. I'll reword the email per @anonymoose2 and your suggestions to be more clear about exactly what is going on.

etnguyen03:
Maybe? Although I still believe that a checkbox is the better way to go (what about if it's a "sticky" counselor meeting was rescheduled - I remember that there are/were some mandatory counselor meetings that were marked as "sticky" throughout the year?)

intranet/templates/eighth/emails/transferred_activity.html
```diff
@@ -0,0 +1,11 @@
+<html><body>
+    {% with description="Room changes" target=info_link name="View status" %}
+        {% include "email_metadata.html" %}
+    {% endwith %}
+
+    <p>
+        You were transferred by the 8th Period Office into {{ dest_act.activity }} on {{ date_str }} during {{ dest_act.block.block_letter }} block. Your previous signup during this block and your signup during {{ source_act.block }} have been removed. Please visit Ion to ensure you are in the correct activities.
```
etnguyen03, line 7:
```suggestion
        You were transferred by the 8th Period Office into {{ dest_act.activity }} on {{ date_str }} during {{ dest_act.block.block_letter }} block. Your previous signup during this block and your signup during {{ source_act.block }} have been removed. Please visit Ion to ensure you are signed up for the activities you are planning to attend.
```

intranet/templates/eighth/emails/transferred_activity.txt
```diff
@@ -0,0 +1,5 @@
+You were transferred by the 8th Period Office into {{ dest_act.activity }} on {{ date_str }} during {{ dest_act.block.block_letter }} block. Your previous signup during this block and your signup during {{ source_act.block }} have been removed. Please visit Ion to ensure you are in the correct activities.
```
etnguyen03, line 1:
```suggestion
You were transferred by the 8th Period Office into {{ dest_act.activity }} on {{ date_str }} during {{ dest_act.block.block_letter }} block. Your previous signup during this block and your signup during {{ source_act.block }} have been removed. Please visit Ion to ensure you are signed up for the activities you are planning to attend.
```

etnguyen03 (changes requested):
The logic looks fine to me; just a change to the email

intranet/apps/eighth/views/admin/scheduling.py
```diff
@@ -402,24 +405,54 @@ def transfer_students_action(request):
     else:
         raise Http404
 
+    if "send_emails" in request.POST:
+        send_emails = True
```
anonymoose2, line 409:
```suggestion
    send_emails = bool(request.POST.get("send_emails"))
```

intranet/templates/eighth/admin/transfer_students.html
```diff
@@ -30,6 +30,16 @@
                 {% endif %}
             {% endcomment %}
         </form>
+    {% elif moved_students %}
+        <p>
+            The following students were already signed up for activities on {{ dest_act.block }}.
+            These signups have been replaced with {{ dest_act }}:
```
anonymoose2, line 36: It might be nice to reiterate whether the students have been notified via email, just for reference. i.e. say either "The students have been notified" or "The students have **not** been notified."

intranet/apps/eighth/views/admin/scheduling.py
```diff
@@ -392,6 +394,7 @@ def transfer_students_action(request):
 
     dest_act = None
     dest_unsignup = False
+    send_emails = False
```
anonymoose2, line 397:
```suggestion
```

anonymoose2 (changes requested):

etnguyen03 (approved):
LGTM

anonymoose2 (approved):
LGTM

PR #1055, opened 2020-10-24
feat(eighth): add filtering of students by activity signup

Laur04:
per 8PO request

### DIFF
--- a/intranet/templates/eighth/admin/multiple_statistics.html
+++ b/intranet/templates/eighth/admin/multiple_statistics.html
@@ -1,0 +1,116 @@
+{% extends "eighth/admin/eighth_admin_page_base.html" %}
+{% load static %}
+{% load pipeline %}
+
+{% block js %}
+    <script src="{% static 'vendor/sortable-0.8.0/js/sortable.min.js' %}"></script>
+    {{ block.super }}
+    <script>
+        $(function() {
+            Sortable.init();
+        });
+    </script>
+{% endblock %}
+
+{% block css %}
+    <link rel="stylesheet" href="{% static 'vendor/sortable-0.8.0/css/sortable-theme-minimal.css' %}">
+    {{ block.super }}
+{% endblock %}
+
+{% block admin_main %}
+    Filter student signups based on activities.
+    <form action="" method="post">
+        {% csrf_token %}
+        {{ form.as_table }}
+        <br>
+        <table>
+            <tr>
+                <th>Lower Signup Limit:</th>
+                <td><input type="text" name="lower" value="{{ lower_absence_limit }}" placeholder="1"></td>
+            </tr>
+            <tr>
+                <th>Upper Signup Limit:</th>
+                <td><input type="text" name="upper" value="{{ upper_absence_limit }}" placeholder="100"></td>
+            </tr>
+            <tr>
+                <th>Start Date:</th>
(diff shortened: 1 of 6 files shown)


intranet/templates/eighth/admin/multiple_statistics.html
```diff
@@ -0,0 +1,115 @@
...
+    <form action="" method="post">
+        {% csrf_token %}
+        {{ form.as_table }}
+        <br>
+        <table>
+            <tr>
+                <th>Lower Limit:</th>
+                <td><input type="text" name="lower" value="{{ lower_absence_limit }}" placeholder="1"></td>
+
+                <td><input type="checkbox" name="freshmen" {% if include_freshmen %}checked{% endif %}> Freshmen</td>
```
etnguyen03, line 31: I think that it would be confusing to include the grade levels as a separate column to the right of the rest of the form; I think that putting the four on one row would be confusing compared to having the grade level checkboxes as a row.

So something like this:

```html
<tr>
    <th>Include:</th>
    <td>
        <input type="checkbox" name="freshmen"  id="freshmen" {% if include_freshmen %}checked{% endif %}>
        <label for="freshmen">Freshmen</label>
        <input type="checkbox" name="sophomores" id="sophomores" {% if include_sophomores %}checked{% endif %}>
        <label for="sophomores">Sophomores</label>
        ...
    </td>
<tr>
```

etnguyen03 (changes requested):
A few things in general:

* The page isn't quite mobile-friendly, but I don't think this is much of a concern?
* It may be beneficial to write unit tests for this change?

intranet/templates/eighth/admin/multiple_statistics.html
Laur04: The reason that I did it this way is it is consistent with the delinquent students view in eighth admin that 8PO mentioned in their email. I agree this probably would look better, but for the sake of consistency? Idk what do you think?

intranet/templates/eighth/admin/multiple_statistics.html
```diff
@@ -0,0 +1,115 @@
...
+
+{% block admin_main %}
+    Filter student signups based on activities.
+    <form action="" method="post">
+        {% csrf_token %}
+        {{ form.as_table }}
+        <br>
+        <table>
+            <tr>
+                <th>Lower Limit:</th>
```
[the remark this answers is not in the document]
Laur04: See comment below about consistency.

intranet/templates/eighth/admin/multiple_statistics.html
etnguyen03: > The reason that I did it this way is it is consistent with the delinquent students view in eighth admin that 8PO mentioned in their email. I agree this probably would look better, but for the sake of consistency? Idk what do you think?

To a new eighth period coordinator/admin I think it _could_ be confusing on the first few tries, but after someone knows how it is done, I don't think it's much of an issue.

Personally I'd change it here and on the "delinquent students view" as well to make it clearer.

Laur04:
1. It doesn't look too bad in mobile...the table overflows a little, but I don't think that's critical.
2. Unit tests added

Everything else addressed.

etnguyen03 (approved):
LGTM

PR #1064, opened 2020-12-09
fix(announcements): expand groups allowed to approve announcements

Laur04:
I've learned to spell announcements correctly!

### DIFF
--- a/intranet/apps/users/models.py
+++ b/intranet/apps/users/models.py
@@ -134,13 +134,40 @@ class UserManager(DjangoUserManager):
         """Returns a ``QuerySet`` containing both teachers and attendance-only users sorted by
         last name, then first name.
 
-        This is used for the announcement request page.
-
         Returns:
             A ``QuerySet`` of teachers sorted by last name, then first name.
 
         """
         return self.get_teachers_attendance_users().order_by("last_name", "first_name")
+
+    def get_approve_announcements_users(self) -> "QuerySet[User]":  # noqa
+        """Returns a ``QuerySet`` containing all users except simple users, tjstar presenters,
+        alumni, service users and students.
+
+        Returns:
+            A ``QuerySet`` of all users except simple users, tjstar presenters, alumni,
+            service users and students.
+
+        """
+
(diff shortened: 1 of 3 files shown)


Laur04:
Codacy's complaint seems to be a non-issue

intranet/apps/users/models.py
```diff
@@ -134,14 +134,41 @@ def get_teachers_attendance_users_sorted(self) -> "QuerySet[User]":  # noqa
...
+        """Returns a ``QuerySet`` containing all users except simple users, tjstar presenters,
+        alumni, service users and students.
+
+        Returns:
+            A ``QuerySet`` of all users except simple users, tjstar presenters, alumni,
+            service users and students.
+
+        """
+
+        users = User.objects.all().exclude(user_type__in=["simple_user", "tjstar_presenter", "alum", "service"])
```
etnguyen03, line 153: It looks like you've forgotten to exclude students.

etnguyen03 (changes requested):
Also, it _might_ be helpful to add unit tests to test this

etnguyen03 (approved):
LGTM

intranet/apps/users/models.py
```diff
@@ -134,14 +134,41 @@ def get_teachers_attendance_users_sorted(self) -> "QuerySet[User]":  # noqa
...
+        """Returns a ``QuerySet`` containing all users except simple users, tjstar presenters,
+        alumni, service users and students.
+
+        Returns:
+            A ``QuerySet`` of all users except simple users, tjstar presenters, alumni,
+            service users and students.
+
+        """
+
+        users = User.objects.all().exclude(user_type__in=["simple_user", "tjstar_presenter", "alum", "service", "student"])
```
theo-o, line 153: I would recommend making this a `filter` rather than `exclude` (in case, future developers want to add additional user types).

Also, `.all()` is not necessary when dealing with `User.objects`.

Laur04: Fixed?

theo-o: Yes.

PR #1067, opened 2020-12-23
feat(auth): warn about incorrect username format

Laur04:
## Proposed changes
- Use javascript and regex to check a user's username upon entry and warn them if it isn't in the correct format 

- Reduce the number of emails about not being to log into Ion

### DIFF
--- a/intranet/static/js/login.js
+++ b/intranet/static/js/login.js
@@ -105,4 +105,19 @@ $(function() {
             });
         }
     });
+
+    function doneTyping () {
+        let re = /^(\d{4})?[a-zA-Z]+\d?$/;
+        if (re.exec($("#id_username").val()) == null) {
+            $("#username-warning").text("Username must be in the format 2016jwoglom for students or jbwoglom for staff.");
+        } else {
+            $("#username-warning").text("");
+        }
+    }
+
+    var typingTimer;
+    $("#id_username").on("keyup", function () {
+        clearTimeout(typingTimer);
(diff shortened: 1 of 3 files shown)


Laur04:
I'm not exactly sure what "Unsafe regex" means...but nothing important dependent on this regex, so I don't think it's an issue?

intranet/static/js/login.js
```diff
@@ -105,4 +105,23 @@ $(function() {
             });
         }
     });
+
+    function doneTyping () {
+        let re = /^(\d{4})?[A-z]+\d?$/;
```
anonymoose2, line 110:
```suggestion
        let re = /^(\d{4})?[a-zA-Z]+\d?$/;
```

`[A-z]` will match everything in `range(ord("A"), ord("z"))`, which includes some extra characters, like `[`, `\`, and `]`. This is probably what Codacy was warning about.

anonymoose2 (changes requested):

intranet/static/js/login.js
Laur04: It didn't solve the issue, but it is fixed. Just scanning through a couple of google searches, it seems that this rule might be complaining about the use of parenthesis because when used as a capturing group, they can provide an avenue for a [ReDoS attack](https://en.wikipedia.org/wiki/ReDoS). I don't believe that is something we need to worry about in this case.

etnguyen03 (approved):
LGTM

anonymoose2 (approved):
LGTM

PR #1069, opened 2020-12-25
fix(polls): change how polls display gender

Laur04:
## Proposed changes
- Change gender display from a back end option to a front end one
- Make the gender option actually work 

Previously there was an GET option `no_gender` that was supposed to display the poll results without the gender counts. However, the columns were still there, it looked cluttered and had IMO unnecessary information. Now, gender stats are always calculated and javascript is used to toggle the tables based on what you want to see.

### DIFF
--- a/intranet/templates/polls/results.html
+++ b/intranet/templates/polls/results.html
@@ -38,10 +38,22 @@
     {{ block.super }}
     <script>
     $(function() {
+        $(".gender-results").hide();
         $("#user-sels").click(function() {
             $(".user-selections").toggle();
         });
-    })
+        $("#gender-toggle").click(function(){
+            if($(this).prop("checked") == true){
+                $(".gender-results").show();
+                $(".grade-header").attr('colSpan', 3);
+                $(".choice-header").attr('rowSpan', 2);
+            } else {
+                $(".gender-results").hide();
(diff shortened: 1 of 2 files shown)


intranet/templates/polls/results.html
```diff
@@ -68,8 +80,8 @@
             &nbsp; <a href="{% url 'poll_csv_results' poll.id %}" class="button small-button print-hide">Download as CSV</a>
         {% endif %}<br>
         <h2>Results: {{ poll }}</h2>
-        <p>&nbsp; {{ poll.get_voted_string }}</p>
-
+        <p>&nbsp; {{ poll.get_voted_string }}</p><br>
+        <p>&nbsp; Show results by gender?<input type="checkbox"></p><br>
```
etnguyen03, line 84: You probably should give this an ID and then reference it in the JS instead of filtering on `[type="checkbox"]` in case we want to add another filter in the future.

etnguyen03 (changes requested):

intranet/templates/polls/results.html
Laur04: Fixed

etnguyen03 (approved):
LGTM

PR #960, opened 2020-02-17
Performance improvements

anonymoose2:
## Proposed changes
- Improve `Announcement.announcementrequest` property -- reduce duplication and never reevaluate
- Prefetch `AnnouncmentRequest`s when listing announcements (they may be queried multiple times)
- Simplify `Announcement.is_visible_requester()` query (WARNING: changes behavior slightly)
- Reuse already-queried `EighthBlock` in signup view
- Don't query unneeded block/waitlist/scheduled activity information
- Fix type annotation
- Improve `EighthActivity.restricted_activities_available_to_user() `
- Fetch user information when listing announcements

## Brief description of rationale
See #958

PR #959, opened 2020-02-17
Fix performance issues

anonymoose2:
## Proposed changes
- Do not needlessly reevaluate `User.recommended_activities `
- Do not load announcementrequest.user unnecessarily
- Consider all waitlists to be empty if `ENABLE_WAITLIST` is `False`
- Make `ENABLE_WAITLIST` default `False` and add warning about performance issues

## Brief description of rationale
See #958

PR #1074, opened 2021-01-22
feat(eighth): enable hybrid

Laur04:
Read [this file](https://github.com/Laur04/ion/blob/hybrid/intranet/apps/eighth/Hybrid-README.rst) for details. 

I'm not by any means done with this; I need to run it through with 8PO and probably write some tests. However, the first iteration of the function is here, so test it out and let me know what you think (after reading the above document).

### DIFF
--- a/intranet/apps/eighth/views/admin/blocks.py
+++ b/intranet/apps/eighth/views/admin/blocks.py
@@ -5,14 +5,16 @@ import re
 from cacheops import invalidate_model
 
 from django import http
+from django.conf import settings
 from django.contrib import messages
 from django.http import HttpResponse
 from django.shortcuts import redirect, render
 from django.urls import reverse
 
 from ....auth.decorators import eighth_admin_required
+from ....users.models import Group
 from ...forms.admin.blocks import BlockForm, QuickBlockForm
-from ...models import EighthBlock, EighthScheduledActivity, EighthSignup
+from ...models import EighthActivity, EighthBlock, EighthScheduledActivity, EighthSignup
 from ..attendance import generate_roster_pdf
 
 logger = logging.getLogger(__name__)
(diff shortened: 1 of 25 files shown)


intranet/apps/eighth/views/admin/blocks.py
```diff
@@ -65,8 +67,59 @@ def add_block_view(request):
 
             invalidate_model(EighthBlock)
 
+    #######
+    if settings.ENABLE_HYBRID_EIGHTH:
+        if request.POST.get("assign_hybrid", "off") == "on":
+            blocks = EighthBlock.objects.filter(date=fmtdate)
+            block_names = {b.block_letter[4:] for b in blocks if "P1" in b.block_letter or "P2" in b.block_letter or "Virt" in b.block_letter}
+            sticky_act = EighthActivity.objects.get_or_create(name="z - Hybrid Sticky", sticky=True, administrative=True)[0]
+            print(block_names)
```
anonymoose2, line 76: Is this old debugging code?

intranet/apps/eighth/views/profile.py
```diff
@@ -86,7 +88,19 @@ def get_profile_context(request, user_id=None, date=None):
     skipped_ahead = False
 
     blocks_all = EighthBlock.objects.order_by("date", "block_letter").filter(date__gte=date)
-    blocks = blocks_all.filter(date__lt=date_end)
+    #######
+    blocks = []
+    if settings.ENABLE_HYBRID_EIGHTH and not request.user.is_eighth_admin:
+        for b in blocks_all:
```
anonymoose2, line 94: In production, you REALLY want to limit the number of DB queries -- the fact that everything's distributed across 5 servers using Docker means that latency is high. When we first moved Ion to Docker, we had massive latency issues, which took some tuning (#959, #960, #964) to work out. (The "SQL" tab of the Django Debug Toolbar is *extremely* useful.)

It's been a while since I wrote Django queries, but I suspect this whole section can be simplified down to 1 or 2 big queries. Something like this (untested; more an outline than a working query): `blocks = blocks_all.filter(eighthscheduledactivity_set__in=EighthScheduledActivity.objects.filter(activity__name="z - Hybrid Sticky").exclude(members_username=request.user.id)).distinct().order_by("date", "block_letter")`

intranet/apps/eighth/views/signup.py
```diff
@@ -133,7 +166,20 @@ def eighth_signup_view(request, block_id=None):
                     # The provided block_id is invalid
                     raise http.Http404 from e
 
-        surrounding_blocks = EighthBlock.objects.get_blocks_this_year()
+        #######
+        surrounding_blocks = []
+        if settings.ENABLE_HYBRID_EIGHTH and not request.user.is_eighth_admin:
+            for b in EighthBlock.objects.get_blocks_this_year():
```
anonymoose2, line 172: This could also likely be simplified.

anonymoose2 (commented):
My main concern here is excessive DB queries causing latency. I've highlighted a few instances of this, but there are of course more.

Laur04:
Yea, those were definitely a concern to fix when I was writing this. Question: does performance significantly degrade when a lookup traverses a lot of `ManyToMany` or `ForiegnKey` relationships or is it more about the number of queries with the distributed setup? I found conflicting opinions online.

anonymoose2:
> Yea, those were definitely a concern to fix when I was writing this. Question: does performance significantly degrade when a lookup traverses a lot of `ManyToMany` or `ForiegnKey` relationships or is it more about the number of queries with the distributed setup? I found conflicting opinions online.

I'm not sure, but I strongly suspect the number of queries is much more important. If you want to take a closer look, you can open up the Django Debug toolbar on Ion (if you're a superuser, just add `?debug=True` to any URL) and see the latencies for yourself.

PR #1080, opened 2021-02-03
build: update dependencies

Laur04:
## Proposed changes
- bump most of our dependencies

Everything is up to date except:
- `celery` because `flower` isn't compatible with the latest release
- `django-redis-cache` because the latest drops support for Django 2.X
- `Django` for obvious reasons

`vine` and `prometheus-client` are also pinned to older releases because of `flower`.


etnguyen03 (changes requested):
When I run I get this

```
/home/vagrant/intranet/intranet/routing.py:29: DeprecationWarning: 
Using ProtocolTypeRouter without an explicit "http" key is deprecated.
Given that you have not passed the "http" you likely should use Django's
get_asgi_application():

    from django.core.asgi import get_asgi_application

    application = ProtocolTypeRouter(
        "http": get_asgi_application()
        # Other protocols here.
    )

  application = ProtocolTypeRouter(
/home/vagrant/.virtualenvs/ion/lib/python3.8/site-packages/channels/routing.py:64: DeprecationWarning: Channel's inbuilt http protocal AsgiHandler is deprecated. Use Django's get_asgi_application() instead.
  self.application_mapping["http"] = AsgiHandler()
```

Maybe we should address this too?

And then there's this from the test suite (see Github Actions output), but it can probably be addressed later

```
Run coverage run ./setup.py test
running test
WARNING: Testing via this command is deprecated and will be removed in a future version. Users looking for a generic test entry point independent of test runner are encouraged to use tox.
```

Laur04:
I addressed everything except the test suite message. That isn't a result of this PR, I believe it's been there for a while. We can address it whenever we move to Django 3, which should probably be sometime this summer.

For that depreciation warning, although `channels` 3.0 _says_ it supports Django 2.2, upon further googling, "support" is not the correct term...more like "don't outright drop". I'll pin the latest 2.X version.

etnguyen03 (approved):
LGTM

PR #1091, opened 2021-02-13
fix(ci): stop push docs from running multiple times

etnguyen03:
## Proposed changes
- Only run the "push docs" step on a given Python version, once, instead of for each Python version

Looking at the latest CIs on `master`, the "push docs" step is being ran on both Python 3.7 and 3.8, and one is failing because they both are trying to push the docs at the same time as the other version, causing a conflict.


Laur04 (approved):
LGTM.

PR #1095, opened 2021-02-21
fix(eighth): move hybrid block signup to celery

Laur04:
## Proposed changes
- Use celery to sign groups up 

Times out in production

### DIFF
--- a/intranet/apps/eighth/Hybrid-README.rst
+++ b/intranet/apps/eighth/Hybrid-README.rst
@@ -29,6 +29,8 @@ Blocks
 
 * In ``intranet/apps/eighth/views/admin/blocks.py``, everything that has been changed is placed between two lines of comments and is behind an if statement checking if the ``ENABLE_HYBRID_EIGHTH`` setting is ``True``. In ``edit_block_view``, a simple warning is given that changing a block name will not change the students that are stickied for that block. Trying to to this here just creates conflicts. 8PO is encouraged to just delete the block and create a new one. In ``add_block_view``, different ``visable_blocks`` are shown and additional context is added. If the "assign_hybrid" checkbox is checked, the appropriate groups will be sticked into the correct admin acitivity according to comments provided throughout and the naming conventions above. As a note, ``attendance_taken`` is set to true, although ``was_absent`` in ``EighthSignup`` is left blank. This is so that 8PO doesn't have to manually take attendance and no one will be marked absent.
 
+* In ``intranet/apps/eighth/tasks.py``, a celery task for hybrid signups is added.
+
 ***********
 Signups
 ***********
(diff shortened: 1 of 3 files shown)


intranet/apps/eighth/Hybrid-README.rst
```diff
@@ -29,6 +29,8 @@ Blocks
 
 * In ``intranet/apps/eighth/views/admin/blocks.py``, everything that has been changed is placed between two lines of comments and is behind an if statement checking if the ``ENABLE_HYBRID_EIGHTH`` setting is ``True``. In ``edit_block_view``, a simple warning is given that changing a block name will not change the students that are stickied for that block. Trying to to this here just creates conflicts. 8PO is encouraged to just delete the block and create a new one. In ``add_block_view``, different ``visable_blocks`` are shown and additional context is added. If the "assign_hybrid" checkbox is checked, the appropriate groups will be sticked into the correct admin acitivity according to comments provided throughout and the naming conventions above. As a note, ``attendance_taken`` is set to true, although ``was_absent`` in ``EighthSignup`` is left blank. This is so that 8PO doesn't have to manually take attendance and no one will be marked absent.
 
+* In ``intranet/apps/eighth/tasks.py``, a celery taask for hybrid signups is added.
```
etnguyen03, line 32:
```suggestion
* In ``intranet/apps/eighth/tasks.py``, a celery task for hybrid signups is added.
```

intranet/apps/eighth/views/admin/blocks.py
```diff
@@ -172,6 +106,75 @@ def add_block_view(request):
...
+            sch_act = EighthScheduledActivity.objects.create(block=b, activity=sticky_act, attendance_taken=True)
+            for g in [Group.objects.get(name="in-person (a-k)")]:
+                for user in g.user_set.all():
+                    try:
+                        sch_act.add_user(user, request=request, force=True, no_after_deadline=True)
+                    except Exception:
+                        failed_users.add(user)
+        celery_logger.debug("Assigned groups successfully.")
+        if failed_users:  # this shouldn't happen because only students should be in the groups, but we don't want to 500
+            celery_logger.debug(
```
etnguyen03, line 138: How do we want to handle errors? Because this likely won't be shown to the user and might go unnoticed in the logs

etnguyen03 (changes requested):

intranet/apps/eighth/views/admin/blocks.py
Laur04: I'm not worried about the `failed_users` error because that would only occur if a teacher or someone had been placed into one of the three groups. They shouldn't be signed up anyway, so I think that error is okay to not return. I supposed I could copy some of this logic into the initial view to output the other error messages as needed (see new force push).

etnguyen03 (approved):
LGTM

PR #1102, opened 2021-02-28
fix(announcements): exclude service and alum from emails

Laur04:

### DIFF
--- a/intranet/apps/announcements/notifications.py
+++ b/intranet/apps/announcements/notifications.py
@@ -12,6 +12,7 @@ from django.contrib.auth import get_user_model
 from django.core import exceptions
 from django.urls import reverse
 
+from ...utils.date import get_senior_graduation_year
 from ..notifications.tasks import email_send_task
 
 logger = logging.getLogger(__name__)
@@ -113,9 +114,18 @@ def announcement_posted_email(request, obj, send_all=False):
     if settings.EMAIL_ANNOUNCEMENTS:
         subject = "Announcement: {}".format(obj.title)
         if send_all:
-            users = get_user_model().objects.all()
+            users = (
+                get_user_model()
+                .objects.filter(user_type="student", graduation_year__gte=get_senior_graduation_year())
+                .union(get_user_model().objects.filter(user_type__in=["teacher", "counselor"]))
+            )
         else:
-            users = get_user_model().objects.filter(receive_news_emails=True)
+            users = (
+                get_user_model()
+                .objects.filter(user_type="student", graduation_year__gte=get_senior_graduation_year())
+                .union(get_user_model().objects.filter(user_type__in=["teacher", "counselor"]))
+                .filter(receive_news_emails=True)
+            )
 
         send_groups = obj.groups.all()
         emails = []


intranet/apps/announcements/notifications.py
```diff
@@ -113,9 +115,13 @@ def announcement_posted_email(request, obj, send_all=False):
     if settings.EMAIL_ANNOUNCEMENTS:
         subject = "Announcement: {}".format(obj.title)
         if send_all:
-            users = get_user_model().objects.all()
+            users = get_user_model().objects.exclude(Q(user_type="service") | Q(graduation_year__gte=get_senior_graduation_year()))
```
etnguyen03, line 118: Do we want to include only students, teachers, and counselors (other user types that might not be supposed to receive these emails, like `user` or `tjstar` users, would still receive them under this)?

Also, this would exclude current seniors. Maybe something like

```suggestion
            users = get_user_model().objects.filter(user_type__in=["student", "teacher", "counselor"], graduation_year__lte=get_senior_graduation_year())
```

intranet/apps/announcements/notifications.py
```diff
@@ -113,9 +115,13 @@ def announcement_posted_email(request, obj, send_all=False):
...
         else:
-            users = get_user_model().objects.filter(receive_news_emails=True)
+            users = (
+                get_user_model()
+                .objects.exclude(Q(user_type="service") | Q(graduation_year__gte=get_senior_graduation_year()))
```
etnguyen03, line 122: Same here - do we want to include only students, teachers, and counselors; instead of excluding service users; and this would exclude current seniors

You _could_ use `graduation_year__gt=get_senior_graduation_year()` if you wanted to use `.exclude()`

etnguyen03 (changes requested):

intranet/apps/announcements/notifications.py
```diff
@@ -113,9 +114,15 @@ def announcement_posted_email(request, obj, send_all=False):
     if settings.EMAIL_ANNOUNCEMENTS:
         subject = "Announcement: {}".format(obj.title)
         if send_all:
-            users = get_user_model().objects.all()
+            users = get_user_model().objects.filter(
+                user_type__in=["student", "teacher", "counselor"], graduation_year__lte=get_senior_graduation_year()
```
etnguyen03, line 118: Wait... sorry (might have been my fault) but wouldn't this exclude all juniors, sophomores, and freshmen?

i.e. if `get_senior_graduation_year()` is 2021, then 2022 is not less than nor equal to 2021, so juniors would be excluded?

```suggestion
                user_type__in=["student", "teacher", "counselor"], graduation_year__gte=get_senior_graduation_year()
```

This would, though, exclude `teacher`s and `counselor`s, because they do not have `graduation_year`s, so maybe we would have to do a union?

etnguyen03 (changes requested):

intranet/apps/announcements/notifications.py
Laur04: So maybe something like:
```
users = get_user_model().objects.filter(user_type="student", graduation_year__gte=get_senior_graduation_year()).union(get_user_model().objects.filter(user_type__in=["teacher", "counselor"]))
```
and then add the extra filter on there for the `else` statement.

etnguyen03: Yeah, I think that will work.

etnguyen03 (approved):
LGTM

PR #1103, opened 2021-03-02
fix(events): fix home template

Laur04:
The edit button doesn't actually show up because context passes `is_events_admin` and this was looking for `events_admin`.


etnguyen03 (approved):
LGTM assuming CI passes

PR #1110, opened 2021-03-12
feat(hybrid): use cache for hybrid groups

Laur04:
## Proposed changes
- Move hybrid groups creation to a different place in the dashboard.

After 8PO tried it out, they asked if it could be moved to a different place that made more sense to them. I looked into it and moving it would allow us to utilize the cache in the same way the rest of the dashboard does.


etnguyen03 (approved):
LGTM but check MM before merging

PR #1115, opened 2021-03-24
fix(hybrid): fix eighth sponsor widget

Laur04:
After attendance is taken, these classes cause the widget to render incorrectly in hybrid mode.


etnguyen03 (approved):
LGTM

PR #1116, opened 2021-03-25
test: add tests for announcements, dataimport, schedule apps

etnguyen03:
## Proposed changes
- Adds unit tests for the `announcements`, `dataimport`, and `schedule` apps.

### DIFF
--- a/intranet/apps/announcements/tests.py
+++ b/intranet/apps/announcements/tests.py
@@ -1,10 +1,13 @@
+import json
+
 from django.contrib.auth import get_user_model
 from django.urls import reverse
+from django.utils import timezone
 
 from ...test.ion_test import IonTestCase
 from ...utils.date import get_senior_graduation_year
 from ..users.models import Group
-from .models import AnnouncementRequest
+from .models import Announcement, AnnouncementRequest
 
 
 class AnnouncementTest(IonTestCase):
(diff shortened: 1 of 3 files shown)


intranet/apps/announcements/tests.py
```diff
@@ -94,3 +105,371 @@ def test_announcement_request(self):
...
+    def test_approve_announcement_view(self):
+        teacher = get_user_model().objects.get_or_create(username="awilliam1", user_type="teacher", first_name="a", last_name="william")[0]
+        user = self.login()
+        request = AnnouncementRequest.objects.create(title="test", content="hello", author="Sysadmins", user=user)
+
+        self.login("awilliam1")
+
+        # The user is not in the "requested_teachers" list, so this should 302 to the index page
+        response = self.client.get(reverse("approve_announcement", kwargs={"req_id": request.id}), follow=True)
+        self.assertEqual(200, response.status_code)
```
Laur04, line 129: I believe you could better test that it really `302`s by first making the request without `follow=True`. So:
```
 response = self.client.get(reverse("approve_announcement", kwargs={"req_id": request.id}))
 self.assertEqual(302, response.status_code)
```
and then what you have here.

Laur04 (changes requested):

intranet/apps/announcements/tests.py
etnguyen03: The check for `response.context["messages"]` (the next line) does not work without `follow=True` and asserting a 200 response, I believe. But I'll try to see.

Laur04: No, that line won't work without `follow=True`. I'm suggesting leaving what you have and add the two lines I wrote above line 128.

etnguyen03: Oh... okay. Makes sense.

Laur04 (approved):
LGTM

PR #1117, opened 2021-03-25
fix(eighth): fix transfer for same block

Laur04:
Currently, when you try to move students from one activity to another when the activities are in the same block, the duplicate signup prevention deletes the `source_act` signup. When the code later tries to change the `scheduled_activity` of all the `source_act` signups, there is nothing there. Essentially, those students just have their signups dropped. 

We can fix this by just ignoring the duplicate signup prevention code when the two blocks are the same. This is okay because there is no way duplicate signups can happen in this case. When the process starts, the student is signed up for exactly one activity (`source_act`) during the block in question and that signup gets changed to `dest_act`.


etnguyen03 (approved):
LGTM, I think

PR #1118, opened 2021-03-27
fix(hybrid): fix future_block out of range error

etnguyen03:
Basically, when a student loads the eighth period signup page, and when hybrid Ion is enabled but there are no future blocks, `future_block[0]` results in an `IndexError`, and the eighth period signup page doesn't load. This commit/PR fixes that.


Laur04 (changes requested):
Code LGTM, but can you rename the commit “fix(hybrid): ...”. That’ll help me track for reverting.

Laur04 (approved):
LGTM

PR #1121, opened 2021-04-06
fix(ci): install pyyaml before building CI YAML

etnguyen03:
## Proposed changes
- Install `pyyaml` before attempting to build the Github Actions `ci.yml` file, which requires PyYAML to be installed.

If `pyyaml` was not installed, then the `ci.yml` file could not be built, so when the CI attempts to verify whether the `ci.yml` file changed at all (as a result of forgetting to run `ci/regen-workflow.py`), nothing happens.


Laur04 (approved):
LGTM

PR #1123, opened 2021-04-15
feat(eighth): automatically sticky withdrawn students

Laur04:
## Proposed changes
- Automatically sticky students in the "Withdrawn from TJ" group into "Z - Withdrawn from TJ" activity when 8PO checks a button.- 

8PO asked for it today when I went to talk to them about the 4 days a week transition.

### DIFF
--- a/intranet/apps/eighth/views/admin/blocks.py
+++ b/intranet/apps/eighth/views/admin/blocks.py
@@ -68,6 +68,25 @@ def add_block_view(request):
 
             invalidate_model(EighthBlock)
 
+        if request.POST.get("assign_withdrawn", "off") == "on":
+            try:
+                grp = Group.objects.get(name="Withdrawn from TJ")
+                act = EighthActivity.objects.get(name="Z - Withdrawn from TJ")
+                for block in EighthBlock.objects.filter(date=fmtdate):
+                    sch_act = EighthScheduledActivity.objects.update_or_create(
+                        block=block,
+                        activity=act,
+                        defaults={"attendance_taken": True, "administrative": True, "sticky": True, "capacity": 100, "restricted": True},
+                    )[0]
+                    for u in grp.user_set.all():
+                        sch_act.add_user(u, request=None, force=True)
+                messages.success(request, "Successfully signed withdrawn students up for the withdrawn activity.")
+            except (Group.DoesNotExist, EighthActivity.DoesNotExist):
+                messages.error(
+                    request,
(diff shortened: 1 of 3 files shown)


intranet/apps/eighth/views/admin/blocks.py
```diff
@@ -68,6 +68,23 @@ def add_block_view(request):
 
             invalidate_model(EighthBlock)
 
+        if request.POST.get("assign_withdrawn", "off") == "on":
+            try:
+                grp = Group.objects.get(name="Withdrawn from TJ")
+                act = EighthActivity.objects.get(name="Z - Withdrawn from TJ")
+                for block in EighthBlock.objects.filter(date=fmtdate):
+                    sch_act = EighthScheduledActivity.objects.get_or_create(
+                        block=block, attendance_taken=True, administrative=True, sticky=True, activity=act, capacity=100, restricted=True
```
etnguyen03, line 77: I think this should be `update_or_create`, to prevent creating a duplicate `EighthScheduledActivity` should one already exist but has (a) modified attribute(s)?

```suggestion
                    sch_act = EighthScheduledActivity.objects.update_or_create(
                        block=block, activity=act,
                        defaults={"attendance_taken": True, "administrative": True, "sticky": True, "capacity": 100, "restricted": True}
```

etnguyen03 (changes requested):
Unit tests would be nice :)

intranet/apps/eighth/views/admin/blocks.py
```diff
@@ -68,6 +68,25 @@ def add_block_view(request):
 
             invalidate_model(EighthBlock)
 
+        if request.POST.get("assign_withdrawn", "off") == "on":
+            try:
+                grp = Group.objects.get(name="Withdrawn from TJ")
```
etnguyen03, line 73: I can't quite remember if this is the case, but does this group have a specific group number?

etnguyen03 (commented):
One thing - but otherwise LGTM

intranet/apps/eighth/views/admin/blocks.py
Laur04: Yes, but I don't know how 8PO handles it. I used the name on both the activity and the group name in case either gets deleted and recreated. We could manually reassign the same ID of course, but I think it is clearer for future devs and 8PO to understand that the name needs to match than the ID?

etnguyen03: Yeah... makes sense.

etnguyen03 (approved):
LGTM

PR #1130, opened 2021-04-30
Character rendering + Pylint fix

Laur04:
## Proposed changes
- Replace backslashes with forward slashes in `title`, `comments` and `admin_comments` fields.
- Make minor formatting changes to align with new pylint standards

This issue was reported by 8PO. Essentially, JSON does a really bad job of handling backslashes with any sort of variability in format. There is no case in entering titles or comments where a backslash is needed -- forward slashes work just as well -- so we can just replace the backslashes.

For pylint, it was complaining about not using `with` to open files in `files.views` and not using `min()` in users.models.


etnguyen03 (approved):
LGTM

PR #1131, opened 2021-05-05
fix(announcements): add missing ul tag on approved posts

etnguyen03:
## Proposed changes
Add the missing `<ul>` tag on the list of approved announcement posts

Without this, the list of approved announcements doesn't line up and looks bad.

image


Laur04 (approved):
Assuming CI passes, LGTM.

PR #1132, opened 2021-05-05
build(deps): bump dependencies

etnguyen03:
## Proposed changes
- Bump dependencies

## Why

CVEs.


Laur04 (approved):
LGTM and this checks out in my dev env, but let's hold the deploy for a safe time just in case.

etnguyen03:
Bumping again for [Django 2.2.22](https://www.djangoproject.com/weblog/2021/may/06/security-releases/)

Even though we don't use Python 3.9, I may as well apply it

PR #1137, opened 2021-05-15
fix(events): delete events without a specified time

Laur04:


etnguyen03 (commented):
Do we think that it's better to generate the timestamp before the event is created, and if creating the timestamp results in a `ValueError`, skip the event instead of creating then deleting the event if it doesn't work?

Laur04:
Yea, probably - just pushed new changes.

etnguyen03 (approved):
LGTM, but see note in ~ion-devs before merging

PR #1140, opened 2021-05-19
feat(eighth): add ability to set 'signup_time' on eighth block creation

### LINKED ISSUE #1097: Add ability to set `EighthBlock` `signup_time` upon creation
## Feature description
On the block creation page, there is no option to set the signup time for blocks. One has to manually go into the edit page for each block and set it.

8PO asked us to do this and it's annoying. The default is 12:40, but they seem to be using differing times.

- This would probably be as easy as adding an <input> to the template and then doing something with that input in the view.

nkenschaft:
closes #1097

The old PR got very messy so I made this new one. It should include all the proposed changes and be ready to merge.

### DIFF
--- a/intranet/apps/eighth/views/admin/blocks.py
+++ b/intranet/apps/eighth/views/admin/blocks.py
@@ -1,3 +1,4 @@
+import datetime
 import logging
 import pickle
 import re
@@ -35,11 +36,13 @@ def add_block_view(request):
 
     date = None
     show_letters = None
+    signup_time = request.POST.get("signup_time") if "signup_time" in request.POST else datetime.time(12, 40)
 
     if "date" in request.GET:
         date = request.GET.get("date")
     if "date" in request.POST:
         date = request.POST.get("date")
+
     title_suffix = ""
     if date:
         date_format = re.compile(r"([0-9]{2})\/([0-9]{2})\/([0-9]{4})")
@@ -57,14 +60,21 @@ def add_block_view(request):
                 if not ltr:
                     continue
                 if ltr not in current_letters:
-                    EighthBlock.objects.create(date=fmtdate, block_letter=ltr)
-                    messages.success(request, "Successfully added {} Block on {}".format(ltr, fmtdate))
+                    EighthBlock.objects.create(date=fmtdate, block_letter=ltr, signup_time=signup_time)
+                    messages.success(
+                        request, "Successfully added {} Block on {} with signups shown to students as closing at {}".format(ltr, fmtdate, signup_time)
+                    )
             for ltr in current_letters:
                 if not ltr:
                     continue
                 if ltr not in letters:
                     EighthBlock.objects.get(date=fmtdate, block_letter=ltr).delete()
                     messages.success(request, "Successfully removed {} Block on {}".format(ltr, fmtdate))
+                else:
+                    blk = EighthBlock.objects.get(date=fmtdate, block_letter=ltr)
+                    blk.signup_time = signup_time
+                    blk.save()
+                    messages.success(request, "Successfully changed the signup time of {} Block on {}".format(ltr, fmtdate))
 
             invalidate_model(EighthBlock)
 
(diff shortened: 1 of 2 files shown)


intranet/templates/eighth/admin/add_block.html
```diff
@@ -46,6 +46,12 @@
                 <input type="text" name="blocks" size="12" maxlength="10">
             </td>
         </tr>
+        <tr>
+            <th><label>Set time:</label></th>
```
Laur04, line 50:
```suggestion
            <th><label>Signup locking time:</label></th>
```

intranet/templates/eighth/admin/add_block.html
```diff
@@ -46,6 +46,12 @@
...
+            <td>
+                <input type="time" name="signup_time" size="12" maxlength="10" value="12:40">
```
Laur04, line 52:
```suggestion
                <input type="time" name="signup_time" size="12" maxlength="10" value="{{ signup_time }}">
```

intranet/apps/eighth/views/admin/blocks.py
```diff
@@ -57,8 +60,10 @@ def add_block_view(request):
...
                 if ltr not in current_letters:
-                    EighthBlock.objects.create(date=fmtdate, block_letter=ltr)
-                    messages.success(request, "Successfully added {} Block on {}".format(ltr, fmtdate))
+                    EighthBlock.objects.create(date=fmtdate, block_letter=ltr, signup_time=signup_time)
+                    messages.success(
+                        request, "Successfully added {} Block on {} with signups shown to students as closing at {}".format(ltr, fmtdate, signup_time)
+                    )
             for ltr in current_letters:
                 if not ltr:
                     continue
```
Laur04, line 69: You should add something like
```
else:
    blk = EighthBlock.objects.get(date=fmtdate, block_letter=ltr)
    blk.signup_time = signup_time
    blk.save()
    messages.success(request, "Successfully changed the signup time on {} Block on {}".format(ltr, fmtdate))
   ```
   under the if statement in line 70. This will allow the signup time to be changed.

File: intranet/apps/eighth/views/admin/blocks.py [hunk 4]
```diff
@@ -122,6 +127,7 @@ def add_block_view(request):
         "date": date,
         "letters": letters,
         "show_letters": show_letters,
+        "signup_time": signup_time,
```
Laur04, line 130:
```suggestion
        "signup_time": str(signup_time)[:5],
```

intranet/apps/eighth/views/admin/blocks.py
```diff
@@ -122,6 +127,7 @@ def add_block_view(request):
         "date": date,
```
Laur04, line 127: Above line 112, add `signup_time = onday[0].signup_time` to pick up the new signup_time.

Laur04 (changes requested):
Okay, there are two things:
1. We want the template to reflect the most recent `signup_time` set, so we need to use the context variable that was passed.
2. Users should be able to update the `signup_time` by hitting "Change".

nkenschaft:
@Laur04 I think I added all the changes you requested, but I'll leave them unresolved so you can verify. Also, if the CI breaks (it's running as of now) I'll of course fix it.

intranet/apps/eighth/views/admin/blocks.py
```diff
@@ -109,6 +119,8 @@ def add_block_view(request):
     #######
     if show_letters:
         onday = EighthBlock.objects.filter(date=fmtdate)
+        if len(onday) > 0:
```
Laur04, line 122:
```suggestion
        if onday:
```

Laur04 (changes requested):
CI is complaining about this one thing, but otherwise LGTM. I'll merge once this is fixed.

intranet/apps/eighth/views/admin/blocks.py
nkenschaft: Done.

Laur04 (approved):
LGTM, pending CI

PR #1141, opened 2021-05-28
fix(eighth): use get_true_rooms when filtering

Laur04:
When using the Room Utilization function in the Eighth Admin dashboard, the logic filters out all room associated with an activity, not just those in use on that particular day. For instance, if Sysadmins was usually held in Room 1 and the Sysadmins `EighthActivity` had Room 1 listed, but then the activity had been overridden to happen in Room 2 at the `ScheduledEighthActivity` level, Room 1 wouldn't show up as open.


etnguyen03 (approved):
LGTM I think

PR #1147, opened 2021-05-30
build: minor ci cleanup

pefoley2:


ci/spec.yml
```diff
@@ -134,19 +134,18 @@ jobs:
...
         run: ./scripts/build_ensure_no_changes.sh ./scripts/build_docs.sh
       - name: Build sources
-        run: ./scripts/build_ensure_no_changes.sh ./scripts/build_sources.sh
+        run: |
+             source venv/bin/activate
+             ./scripts/build_ensure_no_changes.sh ./scripts/build_sources.sh
 
       # Check for changes to CI spec
-      - name: Check for chenges to CI spec
+      - name: Check for changes to CI spec
```
Laur04, line 148: This is already a part of #1148.

ci/spec.yml
```diff
@@ -134,19 +134,18 @@ jobs:
...
-          pip install pyyaml
-          ./scripts/build_ensure_no_changes.sh ./ci/regen-workflow.py
+        run: ./scripts/build_ensure_no_changes.sh ./ci/regen-workflow.py
 
       # Build docs/sources
       - name: Build docs
         run: ./scripts/build_ensure_no_changes.sh ./scripts/build_docs.sh
       - name: Build sources
-        run: ./scripts/build_ensure_no_changes.sh ./scripts/build_sources.sh
+        run: |
```
Laur04, line 143: Doesn't the virtual env already get activated earlier?

ci/spec.yml
```diff
@@ -134,19 +134,18 @@ jobs:
       - *copy_secret
 
       - name: Regenerate workflow YAML file and check for changes
-        run: |
-          set -e
-          pip install pyyaml
-          ./scripts/build_ensure_no_changes.sh ./ci/regen-workflow.py
+        run: ./scripts/build_ensure_no_changes.sh ./ci/regen-workflow.py
```
Laur04, line 137: This is already a part of #1143.

.github/workflows/ci.yml
```diff
@@ -1,3 +1,3 @@
...
...127.0.0.1 -c 'create database ion'"}, {"name": "install Kerberos", "run": "sudo apt install -y krb5-user"}, {"name": "Run tests", "run": "coverage run ./setup.py test"}, {"name": "Migrate database", "run": "coverage run -a ./manage.py migrate"}, {"name": "Collect static files", "run": "coverage run -a ./manage.py collectstatic --noinput -v 0"}, {"name": "Report coverage to Coveralls", "uses": "AndreMiras/coveralls-python-action@develop", "with": {"parallel": true}}, {"name": "Build coverage XML file", "if": "github.repository_owner == 'tjcsl' && github.event_name != 'pull_request'", "run": "coverage xml"}, {"name": "Report coverage to Codacy", "uses": "codacy/codacy-coverage-reporter-action@master", "if": "github.repository_owner == 'tjcsl' && github.event_name != 'pull_request'", "with": {"coverage-reports": "coverage.xml", "project-token": "${{ secrets.CODACY_PROJECT_TOKEN }}"}}]}, "finish_success": {"needs": ["linting", "formatting", "build", "tests"], "runs-on": "ubuntu-latest", "steps": [{"name": "Tell Coveralls that parallel jobs have finished", "uses": "coverallsapp/github-action@master", "with": {"github-token": "${{ secrets.GITHUB_TOKEN }}", "parallel-finished": true}}]}}}
```
Laur04, line 3: See comments on `ci/spec.yml

Ion.egg-info/PKG-INFO
```diff
@@ -1,47 +1,50 @@
...
-        Intranet maintainer emiriti: `Theo Ouzhinski <https://github.com/theo-o>`_ (TJ 2020) & `anonymoose2 <https://github.com/anonymoose2>`_
-        
 Platform: UNKNOWN
 Classifier: Development Status :: 5 - Production/Stable
 Classifier: License :: OSI Approved :: GNU General Public License v2 or later (GPLv2+)
 Classifier: Operating System :: POSIX :: Linux
 Classifier: Programming Language :: Python :: 3.6
 Classifier: Programming Language :: Python :: 3.7
 Classifier: Framework :: Django :: 1.11
+License-File: COPYING
```
Laur04, line 16: Any particular reason for moving this block? Just curious; I'm not super familiar with the convention here, but this part LGTM.

ci/spec.yml
pefoley2: Yup, I figured I'd just rebase whichever one goes in second as otherwise I'd have to muck around w/ downgrading my local setuptools to work around this.

ci/spec.yml
pefoley2: Yup, I'll just rebase whichever one looses the race.

ci/spec.yml
pefoley2: Apparently not?
Or at least when I hacked up build_sources.sh to pip show setuptools, it was the 47 build, not the 57 one installed in the venv.
I'm not sure *why* the venv isn't activated, but it seems that something is missing...

Ion.egg-info/PKG-INFO
pefoley2: It's apparently that way the 2.1 metadata format works, which setuptools 57+ picks by default.

ci/spec.yml
Laur04: Hmm that's really weird. I wouldn't think it would work at all when running CI if the venv wasn't active. So if you run CI without (re?)activating the venv you see a different version of setuptools than if you don't?

pefoley2: yup.
Although now that I look closer, the shebang for manage.py is python3 and setup.py is python, so it might be that the venv only overwrites the former and the latter pull in system python2. Let me try changing that and see if that solves this...

pefoley2: Nope, that wasn't it.
Given that it doesn't seem like venvs are used at all in the official github example, I'm wondering if it's even worth having one here...
ref https://docs.github.com/en/actions/guides/building-and-testing-python

Laur04: Yea, if the build section works without a venv, I don't see anything on casual inspection that would necessitate it.

Laur04 (commented):
Rebase this against `dev`, run `./build_sources.sh`, then LGTM and I'll merge.

pefoley2:
Rebased onto #1153, ptal.

PR #1149, opened 2021-05-30
build: bump celery

pefoley2:
Remove obsolete flower pins, as the newest celery build isn't
compatible.


docs/rtd-requirements.txt
```diff
@@ -32,13 +32,11 @@ django-widget-tweaks==1.4.8
 djangorestframework==3.12.4
 docutils==0.16
 Fabric3==1.14.post1
-flower==0.9.7
 gunicorn==20.1.0
 hiredis==2.0.0
 ipython==7.23.1
 objgraph==3.5.0
 pexpect==4.8.0
-prometheus-client==0.8.0  # pinned for flower
```
Laur04, line 41: Why is this removed? It is used by `flower` and `django-prometheus`.

docs/rtd-requirements.txt
```diff
@@ -32,13 +32,11 @@ django-widget-tweaks==1.4.8
 djangorestframework==3.12.4
 docutils==0.16
 Fabric3==1.14.post1
-flower==0.9.7
```
Laur04, line 35: Why is this removed?

docs/rtd-requirements.txt
```diff
@@ -51,7 +49,6 @@ sentry-sdk==1.0.0
 setuptools-git==1.2
 Sphinx==3.5.4
 sphinx-bootstrap-theme==0.7.1
-vine==1.3.0  # pinned for flower
```
Laur04, line 54: Why is this removed? It is used by `flower`.

Ion.egg-info/requires.txt
```diff
@@ -51,7 +49,6 @@ sentry-sdk==1.0.0
 setuptools-git==1.2
 Sphinx==3.5.4
 sphinx-bootstrap-theme==0.7.1
-vine==1.3.0
```
Laur04, line 54: If we can bump `flower`, we can just bump this to be compatible.

docs/rtd-requirements.txt
Laur04: [This](https://github.com/mher/flower/issues/1029#issuecomment-835486909) seems to suggest that we might have to modify our production setup slightly, but `flower` + `celery` should still work.

pefoley2: It looks like there hasn't actually been a new build cut that supports celery 5.0:
https://github.com/mher/flower/issues/1109
So it looks like it makes sense to either wait for that to be resolved or avoid bumping celery in the meantime.
That said, I can't actually see where flower is used in ion...

Laur04: I'm inclined to wait. This looks like a pretty active area of discussion and `flower` seems to be maintained. We use `flower` in the production setup for monitoring `celery`. I'm going to leave this PR open as a reminder to check back for a new `flower` release.

pefoley2: SGTM, I'll sit on this for now.

Laur04:
This seems to have been released, so I'll include this in my next round of dependency bumps.

Laur04:
@pefoley2 Rebase this against the latest `dev`. We can bump `celery`, `certifi`, `flower`, `prometheus-client` and `vine`, but we can't downgrade `channels`.

Laur04:
There are breaking changes going from 4 to 5. Give me a while today to investigate.

Laur04:
Okay @pefoley2 it looks mostly good. Two things:

1. Ion is Python 3.8 in production, so the failing test there isn't a huge concern. However, I want to keep testing 3.7 for now, so could you please do as suggested [here](https://github.com/celery/kombu/issues/1339#issuecomment-887336714)? That thread also gives context to why this this failing and why it isn't a huge concern for us.
2. Celery 5.0 introduces breaking changes to the CLI. That doesn't have the potential to affect anything but `intranet/scripts/restart_celery.sh` which invokes celery from the command line. It *looks* fine, but I'd like a second set of eyes. Details on the breaking changes are [here](https://docs.celeryproject.org/en/stable/history/whatsnew-5.0.html#v500-important)

Laur04 (changes requested):
See latest comment

Laur04:
Addressed with other dependency bumps in 32495cad21ff06dd0105d58b22f25e757485dab2. Closing

pefoley2:
Hm, that change appears to have been reverted?
https://github.com/tjcsl/ion/commit/3d4dc846f22741a88efa90775e7c5273fc05c722

Laur04:
Yea, we had some issues going to production with the new command line invocation and some dependencies. I'm going to reapply all of this when I get a free weekend.

PR #1150, opened 2021-05-31
Bump deps and resolve django3 incompatibilities 

pefoley2:


Laur04 (approved):
LGTM...the plan is to officially make the upgrade to Django 3 this summer. I'd rather get #1147 merged first and then rebase against that because this contains the pkg-info changes.

Laur04 (changes requested):
This needs to be rebased against `dev` which now includes #1153

PR #1151, opened 2021-06-02
fix(vagrant): remove unnecessary VirtualBox GUI

nkenschaft:
This has been repeatedly mentioned and no one got around to doing it. As far as I can tell, the VirtualBox GUI serves no purpose for the Ion development environment, so it seems best to just get rid of it.


Laur04 (approved):
LGTM

PR #1153, opened 2021-06-10
build: stop using venv

pefoley2:
This lets ./setup.py egg_info pick up the newer setuptools we installed
so that we get the 2.1 metadata format.


Laur04 (approved):
LGTM. This by necessity covers most of #1147, so I'll merge it and you can change  #1147 accordingly.

PR #1158, opened 2021-06-12
build: django3 compat

pefoley2:
Fix a bunch of warnings that show up when running under Django 3.2


intranet/apps/bus/migrations/0005_alter_route_options.py
```diff
@@ -0,0 +1,17 @@
...
+    dependencies = [
+        ('bus', '0004_auto_20180117_1232'),
+    ]
+
+    operations = [
+        migrations.AlterModelOptions(
+            name='route',
+            options={'ordering': ['route_name']},
+        ),
+    ]
```
Laur04, line 17: Seems strange that there is a migration when `bus/models.py` doesn't seem to have changed.

pefoley2: The migration is from a previous change, I forgot to run makemigrations when doing https://github.com/tjcsl/ion/commit/900edd5aa848148fdf92551dadb3da0f13773c57#diff-488cc7ad7f58d5a2e8f027e857ea887909a1b0df6c04dce0cb76fc0cabf46e15

Laur04 (approved):
LGTM

PR #1161, opened 2021-07-05
feat(emailfwd): add senior email forwards command

smayya337:
Closes tjcsl#1152.

## Proposed changes
- Add `intranet/apps/emailfwd/management/commands/get_senior_forwards.py`, a management command which retrieves and prints senior mail forwards.
- Start work on a test for the management command. This still needs some work - I'm a bit unsure on how to retrieve the output from stdout.

This implementation is based on @etnguyen03's suggestion of outputting formatted strings to stdout.

### DIFF
--- a/intranet/apps/emailfwd/management/commands/get_senior_forwards.py
+++ b/intranet/apps/emailfwd/management/commands/get_senior_forwards.py
@@ -1,0 +1,13 @@
+from django.core.management.base import BaseCommand
+
+from .....utils.date import get_senior_graduation_year
+from ....emailfwd.models import SeniorEmailForward
+
+
+class Command(BaseCommand):
+    help = "Gets all senior email forwards"
+
+    def handle(self, *args, **options):
+        forwards = SeniorEmailForward.objects.filter(user__graduation_year=get_senior_graduation_year(), user__user_type="student")
+        for forward in forwards:
+            self.stdout.write("%s:\t\t%s" % (forward.user, forward.email))
--- a/intranet/apps/emailfwd/tests.py
+++ b/intranet/apps/emailfwd/tests.py
@@ -1,4 +1,7 @@
+from io import StringIO
+
 from django.contrib.auth import get_user_model
+from django.core.management import call_command
 from django.urls import reverse
 
 from ...test.ion_test import IonTestCase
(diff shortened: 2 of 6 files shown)


intranet/apps/emailfwd/management/commands/get_senior_forwards.py
```diff
@@ -0,0 +1,12 @@
+from django.core.management.base import BaseCommand
+
+from ....emailfwd.models import SeniorEmailForward
+
+
+class Command(BaseCommand):
+    help = "Gets all senior email forwards"
+
+    def handle(self, *args, **options):
+        forwards = [f for f in SeniorEmailForward.objects.all() if f.user.is_senior]
```
Laur04, line 10:
```suggestion
        forwards = SeniorEmailForward.objects.filter(user__is_senior=True)
```

intranet/apps/emailfwd/tests.py
```diff
@@ -31,3 +31,12 @@ def test_email_fwd(self):
...
         self.assertEqual(1, SeniorEmailForward.objects.filter(user=user, email="nonexistent@tjhsst.edu").count())
         self.assertIn("Error adding forwarding address.", list(map(str, list(response.context["messages"]))))
+
+
+class GetSeniorForwardsTest(IonTestCase):
+    def test_getting_forwards(self):
+        """Test ability to retrieve senior forwards."""
+        user = get_user_model().objects.get_or_create(username="awilliam", graduation_year=get_senior_graduation_year(), user_type="student")[0]
+        self.login()
+        # need to figure out how to read the data, which goes to stdout
```
Laur04, line 41: It might be easiest to log this information in the view and then use tests to check for the logs instead of checking stdout. The other option is to have this command save to a file (which might actually be better for our use case) and then it's fairly easy to test the file output.

Laur04 (changes requested):
In addition to all of this, you need to get CI passing.

intranet/apps/emailfwd/management/commands/get_senior_forwards.py
smayya337: Maybe it's just me, but when I use that I get `django.core.exceptions.FieldError: Related Field got invalid lookup: is_senior` - my understanding is that Django can't `filter` by properties like `is_senior` since they're not actually in SQL, which is why I used a list comprehension instead.

intranet/apps/emailfwd/tests.py
smayya337: Doing some Googling, I came across [`call_command`](https://docs.djangoproject.com/en/2.2/ref/django-admin/#running-management-commands-from-your-code), which seems to be useful for this stuff (and it has `stdout` redirection!) I'm not sure if it's the cleanest solution, but I suppose it's worth a try.

intranet/apps/emailfwd/management/commands/get_senior_forwards.py
Laur04: Ah no you're right -- `is_senior` isn't a database field. You could do `graduation_year`=`get_senior_graduation_year()`, which *should* be *slightly* more memory efficient, but either way is probably fine.

intranet/apps/emailfwd/tests.py
Laur04: That sounds good.

smayya337:
OK, the `build` error was me forgetting to run `build_sources.sh` and the `tests` one was the result of a typo, but what happened with `linting`?

Laur04:
Revert the change to `Ion.egg-info/PKG-INFO`...that's caused by an outdated dev dependency locally. As far as linting goes, sometimes new checks are introduced to pylint and we have to update stuff. I'll fix it in a minute, don't worry about it for now.

smayya337:
> Revert the change to `Ion.egg-info/PKG-INFO`...that's caused by an outdated dev dependency locally.

Do I need to do anything to bring the dependency up-to-date?

Also, it appears that the test I wrote works this time, so that's good.

Laur04 (changes requested):
I don't think you reverted the file...it's still showing changes.

smayya337:
That was weird... somehow a space went missing in there.

Laur04 (changes requested):
Squash commits, otherwise LGTM

smayya337:
Commits squashed, but I don't have merge powers so you'll have to do that part

smayya337:
Ugh, I had the wrong config option set...

Laur04 (approved):
LGTM

PR #1162, opened 2021-07-06
docs(testing): update testing docs

Laur04:


docs/developing/testing.rst
```diff
@@ -2,33 +2,27 @@
...
-Testing functionality that is useful for multiple tests can be found in ``intranet/test``.
+For most modules, the unit tests go in ``intranet/apps/<module>/tests.py``. Currently, the sole exception is ``eighth``, where tests are broken out into several different files under ``intranet/apps/eighth/tests/``. Testing functionality that is useful for multiple tests can be found in ``intranet/test``.
 
 Running Tests
 =============
 
-To actually execute tests, run ``./setup.py test``.
-Note that this deletes and re-creates the db from scratch each time,
-so you may want to pass the ``-k`` option when developing tests as it significantly reduces run-time.
+To actually execute tests, run ``pytest`` inside your vagrant environmnet. If you want to only test a specific apps, you can run ``pytest <apps_name>``. Do note that this deletes and re-creates the db from scratch each time, so you may want to pass the ``-k`` option when developing tests as it significantly reduces run-time.
```
smayya337, line 15:
```suggestion
To actually execute tests, run ``pytest`` inside your vagrant environment. If you want to only test a specific app, you can run ``pytest <apps_name>``. Do note that this deletes and re-creates the db from scratch each time, so you may want to pass the ``-k`` option when developing tests as it significantly reduces run-time.
```

docs/developing/testing.rst
```diff
@@ -39,9 +33,24 @@ Here is an bare-bones example of the basic layout for a test:
       # Put your tests here
       self.assertEqual(1, 1)
 
+Every test should include comments that explain, almost in narrative form, what the test is doing and what the expected results are.
+
+Generally, there are two ways that you test Ion's code. These all not comprehensive, but should work for most cases. The first is calling methods directly; the second is making a GET or POST request to the view that you are interested in testing. The best tests utilize both. After you do either/both of these to call the code, you should use assertions to see if the code behaved as expected. A list of assertion options can be found `here <https://docs.python.org/3/library/unittest`_.html#assert-methods>`. A useful tool for making requests to the view is ``self.client``. More documentation on that can be found `here <https://docs.djangoproject.com/en/3.2/topics/testing/tools/>`_. Alternatively, just search through Ion's testing files for examples of using `self.client`.
```
smayya337, line 38:
```suggestion
Generally, there are two ways that you test Ion's code. These are not comprehensive, but should work for most cases. The first is calling methods directly; the second is making a GET or POST request to the view that you are interested in testing. The best tests utilize both. After you do either/both of these to call the code, you should use assertions to see if the code behaved as expected. A list of assertion options can be found `here <https://docs.python.org/3/library/unittest`_.html#assert-methods>`. A useful tool for making requests to the view is ``self.client``. More documentation on that can be found `here <https://docs.djangoproject.com/en/3.2/topics/testing/tools/>`_. Alternatively, just search through Ion's testing files for examples of using `self.client`.
```

docs/developing/testing.rst
```diff
@@ -39,9 +33,24 @@ Here is an bare-bones example of the basic layout for a test:
...
+
+Good Testing Examples
+=====================
+
+Ion has lots of tests; the following are examples of very well written ones. Note the abundance of comments and the attention to detail. Test everything.
```
smayya337, line 43:
```suggestion
Ion has lots of tests; the following are examples of very well-written ones. Note the abundance of comments and the attention to detail. Test everything.

- `Polls <https://github.com/tjcsl/ion/blob/dev/intranet/apps/polls/tests.py>`_ - Note the thoroughness of the checks; assertions are made even when nothing should have changed.
- `Emailfwd <https://github.com/tjcsl/ion/blob/dev/intranet/apps/emailfwd/tests.py>`_ - Note the use of `self.client` and checking `messages` generated by the page.
- `Eighth <https://github.com/tjcsl/ion/tree/dev/intranet/apps/eighth/tests>`_ - If a testing method has been used in Ion, it's in ``eighth``. This is the most complicated and also the most thorough set of tests that Ion has.
```
`If a testing method has been used in Ion` might need a little bit of rewording - it comes across as a touch confusing.

smayya337 (changes requested):
These are all very small changes, because the rest of it is pretty good!

smayya337 (approved):
LGTM!

etnguyen03 (approved):
LGTM too

PR #1163, opened 2021-07-06
fix(ci): fix linting and tests

Laur04:

### DIFF
--- a/intranet/apps/bus/management/commands/import_routes.py
+++ b/intranet/apps/bus/management/commands/import_routes.py
@@ -9,12 +9,13 @@ class Command(BaseCommand):
     help = "Import Routes from routes.csv"
 
     def handle(self, *args, **options):
-        reader = csv.reader(open("routes.csv", "r"))
-        print("Deleting preexisting buses")
-        for row in reader:
-            (name,) = row
-            _, created = Route.objects.get_or_create(route_name=name)
-            if created:
-                print("Created route {}".format(name))
-            else:
-                print("Route {} already exists")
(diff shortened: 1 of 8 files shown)


Laur04:
Codacy can be ignored

intranet/apps/bus/management/commands/import_routes.py
```diff
@@ -1,3 +1,4 @@
+# pylint: disable=consider-using-with
```
etnguyen03, line 1: Instead of disabling `consider-using-with` could we use

```python
with open(...) as ...:
```

?

Laur04: So `with csv.reader(open("routes.csv", "r")) as reader:` doesn't address this concern; it is worried about the `open("routes.csv", "r")` in the middle. We'd have to do something like:
```
with open("routes.csv", "r") as f:
    reader = csv.reader(f)
```
I didn't think that was very clean, but we could do that.

intranet/apps/users/views.py
```diff
@@ -1,3 +1,4 @@
+# pylint: disable=consider-using-with
```
etnguyen03, line 1: Do we want to change this file too?

etnguyen03 (approved):
Mostly LGTM but one thing

intranet/apps/users/views.py
Laur04: No, the logic wouldn't work very well in that file. It's not a straightforward conversion there.

PR #1166, opened 2021-07-31
feat(polls): add error message to modifying polls after end date

### LINKED ISSUE #1128: Display an error message when trying to edit a poll after it has closed
This issue is being created for training purposes. PLEASE DO NOT SUBMIT A PULL REQUEST FOR THIS ISSUE UNLESS YOU ARE EXPLICITLY AUTHORIZED TO DO SO.

## Steps to reproduce
1. As a polls admin, create a poll with the close time set to sometime in the past.
2. After clicking "Show All" on the polls home page, attempt to edit the poll you have just created, which should be closed.

## Expected behavior
Ion should redirect back to the page the request was sent from, either `/polls` or `/polls?show_all=1` with a message at the bottom that informs the user that they cannot edit a poll that has been closed.

## Actual behavior
Ion returns to `/polls` with no error message.

pizzalemon:
Closes #1128

Adds an error message when a user tries to edit their poll after the end date.

### DIFF
--- a/intranet/apps/polls/views.py
+++ b/intranet/apps/polls/views.py
@@ -11,6 +11,7 @@ from django.core.serializers import serialize
 from django.db import transaction
 from django.db.models import Q
 from django.shortcuts import get_object_or_404, redirect, render
+from django.urls import reverse
 from django.utils import timezone
 
 from ...utils.date import get_senior_graduation_year
@@ -409,7 +410,12 @@ def modify_poll_view(request, poll_id):
     poll = get_object_or_404(Poll, id=poll_id)
 
     if not poll.before_end_time():
-        return redirect("polls")
+        messages.error(request, "Closed polls may not be edited!")
+
+        if "show_all" in request.GET:
+            return redirect(reverse("polls") + "?show_all=1")
+        else:
+            return redirect("polls")
 
     if request.method == "POST":
         form = PollForm(data=request.POST, instance=poll)
--- a/intranet/templates/polls/poll.html
+++ b/intranet/templates/polls/poll.html
@@ -17,7 +17,7 @@
         {% if not poll.visible %} (HIDDEN){% endif %}
         {% if is_polls_admin %}
             <div class="poll-icon-wrapper">
-                <a href="{% url 'modify_poll' poll.id %}">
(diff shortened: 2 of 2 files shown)


Laur04 (changes requested):
What you have looks good, but as stated in the issue, the user should be redirected back to the page they came from. If they came from the page with ?show_all=True set, they should be redirected back to it.

smayya337:
Also, don't worry about the CI failures - they appear to be unrelated to your work. We'll take a look at them later.

pizzalemon:
I added another commit which should do that. I made it so it should go back to polls/ in case it's from there (and of course polls?show_all=1), but as far as I can tell, you can't access an old poll unless you have show all on, so the request would always be from polls?show_all=1 anyways, but the way you phrased it in the issue made it sound like I should plan for it being from polls/, so I made it like that in the commit.

Laur04 (changes requested):
squash your commits into one commit with the commit message of the first.

Laur04:
You also need to fix the CI for formatting and linting. Don't worry about tests.

pizzalemon:
It should be good I think?

intranet/apps/polls/views.py
```diff
@@ -409,7 +410,12 @@ def modify_poll_view(request, poll_id):
     poll = get_object_or_404(Poll, id=poll_id)
 
     if not poll.before_end_time():
-        return redirect("polls")
+        messages.error(request, "Closed polls may not be edited!")
+
+        if "show_all" in request.GET:
+            return redirect(reverse("polls") + "?show_all=1")
```
Laur04, line 416:
```suggestion
            return redirect(reverse("polls", kwargs={"show_all":1}))
```

Laur04 (changes requested):
Otherwise, LGTM. The tests will pass tomorrow.

Laur04 (changes requested):
Okay, I promise that this is the last thing, but put it back to how it was before I suggested a change please. Sorry, that's a syntax for a different version that actually breaks here. Didn't think about that until I actually tested it and it broke unfortunately. After that, I'll run the CI tomorrow and then merge.

Laur04 (approved):
LGTM, pending CI. Congrats on your first contribution.

PR #1170, opened 2021-08-12
build(deps): update to django3

Laur04:
`django-prometheus`, `django-cacheops`, `django-debug-toolbar` and `django-oauth-toolkit` haven't updated to remove some depreciation, but that doesn't actually become an issue until django 4. Because of this however, we have to mute the depreciation warnings in pytest.

I went over the changelogs pretty thoroughly and I'd like to put this in this weekend, after deploying it to `ion-dev`, but please see if I've missed anything.

E: to be clear, still very much testing this

### DIFF
--- a/Ion.egg-info/requires.txt
+++ b/Ion.egg-info/requires.txt
@@ -11,7 +11,7 @@ channels-redis==3.2.0
 contextlib2==0.6.0.post1
 cryptography==3.4.7
 decorator==5.0.9
-Django==2.2.24
+Django==3.2.6
 django-cacheops==6.0
 django-cors-headers==3.7.0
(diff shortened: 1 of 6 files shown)


Laur04:
I don't know why codacy is complaining...that didn't exist in django 2, but it certainly does in django 3.

etnguyen03 (approved):
Huh. I'm not sure about the Codacy error, but the functionality LGTM... but you should probably test thoroughly in `dev` before deploying.

Laur04:
After killing this in `ion-dev` all morning, I haven't run into any errors. I'm going to merge and deploy and then beat it up a little in production to to be extra sure.

PR #1176, opened 2021-09-05
feat(eighth): add functionality to delete empty groups

### LINKED ISSUE #952: Add functionality to delete empty groups
## Feature description
Add a button in the group admin section of the Eighth Admin page to redirect to a page to delete empty groups. This page would contain the listing of groups with 0 members to delete and checkboxes to check as needed.

In production, a variety of groups have 0 members. In most of these cases, these are obsolete groups that don't need to exist.  We should have functionality for appropriate admins to delete these groups.

-

Laur04:
closes #952


smayya337:
Is it possible to write a test here? Maybe you could have it create an empty group, then try to delete and verify that a) it lists the group as available for deletion, and b) actually deletes the group on `POST`.

PR #1178, opened 2021-09-09
feat(eighth): add list sponsors page

### LINKED ISSUE #674: Add list sponsors page to eighth admin

Laur04:
closes #674


smayya337:
This is also another spot where it might be useful to write a test - e.g. add a test user as the sponsor of a test activity then verify that the list of activities they're sponsoring includes this activity. Also, I'm not too familiar with how exactly this stuff is organized, but is there a reason that the CSS is in the view rather than in a separate file or something?

Laur04:
`eighth` is just so extensive that for little css blocks, it's easier to keep everything straight when the css is in the template.

PR #1179, opened 2021-09-09
build(deps): bump pillow

Laur04:
see dependabot


smayya337:
Can't test right now, but I don't think (?) there should be any issues with this dependency bump.

PR #1186, opened 2021-10-30
Create dependabot.yml

pefoley2:
Enable automatic dependency PRs


Laur04:
See https://github.com/tjcsl/ion/pull/1188

pefoley2:
Ah, you've got the v1 config file.
I missed that :|
Lemme re-spin this.

PR #1204, opened 2021-11-30
feat(users): support non-binary genders

### LINKED ISSUE #1174: feature(users): support non-binary genders
## Feature description
This is not a major priority, but currently, the `User.gender` field is set using a boolean - `True` for male, `False` for female. This becomes a problem when non-binary individuals are added to the mix, since that by definition is neither `True` nor `False`.

TJ has a non-negligible non-binary population.

Could we change `User.gender` to be a string instead of a boolean?

shahsalonik:
closes #1174 

## Proposed changes
- Change User.gender to a string
- Add an "is_nonbinary" method

TJ has a large non-binary population; this change could help with inclusivity.


Laur04 (changes requested):
Overall, pretty solid, but there are a few things to sort out.
1. You have missed some places where gender is referenced. For instance, in `intranet/templates/users/profile.html` on line 167, `is_male` is called and if it is `False`, gender is assumed to be female. I would suggest using a search feature to search the rest of the codebase for calls to `is_male` or `is_female` or references to "gender" or "sex" in order to catch all of the places where we might call those methods. Check the API too - it returns gender information.
2. You have some failing tests. Each test will tell you why it failed and you just have to fix that.
3. We need to make a plan to migrate our database (which currently stores boolean values) to this new schema that uses strings. I usually accomplish this with a simple Python script that runs through Django's shell. You can get there by running `python manage.py shell_plus` inside your dev environment and test out a script that will change all Boolean data in the gender field to strings. Let me know what you are thinking about doing and I'll work with you to get that into production.

Laur04:
One other thing that you don't have to worry about yet, but we will want to squash all of these commits down to one commit that follows our [commit naming guidelines](https://github.com/tjcsl/ion/blob/master/docs/developing/howto.rst#formatting-commit-messages). Play around with `git rebase -i` to accomplish this.

shahsalonik:
Sounds good! I’ll keep working on it this week.

PR #1236, opened 2022-01-28
feat(users): support non-binary genders

### LINKED ISSUE #1174: feature(users): support non-binary genders

shahsalonik:
closes #1174 
## Proposed changes
- Change User.gender to a string
- Add an "is_nonbinary" method

TJ has a large non-binary population; this change could help with inclusivity.


shahsalonik:
I think I'm done with this.

Laur04 (changes requested):
This is good in terms of model changes, but there are many other places throughout Ion's codebase where gender is references in templates, views and management commands. I'd Ctrl+F through the entire codebase to find references to gender and make sure they fit with this new variable type. Installing Visual Studio would probably be helpful.

Laur04:
What's the status here?

Laur04 (changes requested):
Let's talk about how best to migrate this change.

intranet/apps/users/models.py
```diff
@@ -562,33 +569,43 @@ def grade_number(self) -> int:
...
-        """Returns "Male" if this user is male, "Female" otherwise.
+        """Returns the gender of this user (male, female, or non-binary).
 
         Returns:
-            "Male" if this user is male, "Female" otherwise.
+            The gender of this user (male, female, or non-binary).
 
         """
-        return "Male" if self.is_male else "Female"
+        return self.gender
```
Rushilwiz, line 578:
```python
return self.gender or ""
```
In case gender is `None`. I think the serializer is the only thing that uses this field.

PR #1247, opened 2022-02-26
build(deps): Update some packages

pefoley2:


Ion.egg-info/requires.txt
```diff
@@ -40,21 +40,21 @@ objgraph==3.5.0
...
-sphinx-bootstrap-theme==0.8.0
+Sphinx==4.4.0
+sphinx-bootstrap-theme==0.8.1
 vine==5.0.0
 asgiref>=3.3.4
 pillow>=9.0.0
 twisted>=21.7.0
 
 [:python_version < "3.8"]
-importlib_metadata==3.6
```
Laur04, line 60: This is to prevent `tests` from failing on `python3.7`, so we want to keep it pinned here.

Laur04 (changes requested):
Otherwise, LGTM. Will merge after 8th periods this week.

PR #1269, opened 2022-03-29
feat: it's Ion not ION

Laur04:


smayya337:
Does this run every time a page is loaded? If so, that's going to infuriate a *lot* of people.

Laur04:
Just whenever the dashboard is loaded. It's not working right now for some reason. The messages will expire after about 10 seconds.

PR #1299, opened 2022-05-13
feat(dev): docker dev env fixes

NotFish232:
Updated dockerfile with libmagic to fix intranet dependency. Update dockker compose to expose redis port to containers instead of host.


Laur04:
Squash these commits and name them appropriately please

Laur04 (changes requested):
See above

Laur04 (changes requested):
Rebase this against the latest `dev`. I think `libmagic` already got fixed.

PR #1303, opened 2022-05-19
fix(events): make attend function work

alanzhu1:
## Proposed changes
- Makes the attend/don't attend event button work.
- The attend button is not part of the attend event form, but is a separate `a` element, which is why it does not work.
- Uses JavaScript to submit the attend form when the attend button is clicked, fixing the issue. 
- Also makes the attend button show up on the dashboard.

- Currently, the attend event button doesn't do anything when clicked.
- See, for example, https://ion.tjhsst.edu/events/784
- The button also doesn't currently show up on the dashboard. Allowing people to mark as attending from the announcement post increases visibility.
- In the future, might want to look into making the attend button `<input type="submit">` to eliminate need for JS. However, this could create formatting issues which is why I chose to keep it where it is for now.

### DIFF
--- a/Ion.egg-info/SOURCES.txt
+++ b/Ion.egg-info/SOURCES.txt
@@ -1006,6 +1006,7 @@ intranet/static/js/busdriver.js
 intranet/static/js/common.header.js
 intranet/static/js/common.js
 intranet/static/js/common.nav.js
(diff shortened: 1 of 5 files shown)


Laur04 (approved):
LGTM

Issue #1211, opened 2021-12-23
Pull status information from CSL Stausfy


Laur04:
## Feature description
In addition to pulling FCPS emergency information, we should pull from https://status.tjhsst.edu.

## Rationale for adding feature
Although sometimes issues result in Ion being down altogether, many have to do with other issues of the lab. Status isn't necessarily widely referenced by the student body, but Ion is and that makes it a better place for centralizing this information.

## Suggested implementation(s) (if applicable)
- The framework from doing this is already present in `emerg`, it just needs to be slightly modified to grab info from Status as well.
- In addition to the celery task that pulls emergency information, there might be a reason to allow administrators to manually trigger an update so that information stays up to date.

dianalin2:
Hi, can I take this one?

Laur04:
Yep

smayya337:
Oops, I didn't realize which issue this was - @dianalin2 there's one problem to solve before you do this issue, which is that status.tjhsst.edu is broken and we need to replace it. (You're welcome to help me find a new status page that'll play nice with info pulling, though!)

alanzhu0:
Update: we now need to pull from CSL BetterUptime. 
https://betterstack.com/docs/uptime/api/getting-started-with-uptime-api/

Also, FCPS seems to have changed their alert format - we need to fix that too.

PR #1304, opened 2022-05-20
feat: add csl statusfy integration in emergency. refs #1211

dianalin2:


intranet/apps/emerg/views.py
```diff
@@ -68,12 +68,30 @@ def check_emerg():
...
+        r = requests.get(settings.CSL_STATUS_PAGE, timeout=timeout)
+    except requests.exceptions.Timeout:
+        pass
+
+    try:
+        csl_status = json.loads(r.text)
+    except json.decoder.JSONDecodeError:
+        return False, None
+
+    for system in csl_status['systems']:
```
smayya337, line 83: I think that with the code that's currently here, the status isn't updated if there's a  CSL issue, so you'll never see a CSL issue unless there's an FCPS emergency announcement as well. Having `status` update whenever there's a CSL issue will fix that. Also, we probably don't want informational updates to be in the yellow (e.g. current Borg cluster maintenance, so we can filter that out), and we shouldn't have multiple instances of the same thing in the `announcements` list (the JSON has the same issue under each affected system).

With all that said, here's how I worked around all of this:
```py
status = False
    for system in csl_status['systems']:
        if system['status'] != 'ok':
            status = True
            issues = system['unresolvedIssues']
            for issue in issues:
                desc = requests.get(issue['permalink']).text
                soup = BeautifulSoup(desc, "html.parser")
                desc = soup.find_all('p')[1].text
                val = {"title": issue['title'], "body": desc}
                if val not in announcements and issue['severity'] != 'notice':
                    announcements.append(val)
```

smayya337 (changes requested):
I noticed one issue that prevented the CSL status updates from showing up on my end, as well as a few places to tweak how everything is displayed. Besides that, the only other thing I can think of is maybe removing the "-FCPS" in the bottom right corner of the announcement area, and this should be good (though I defer to @Laur04 there).

Laur04 (changes requested):
Changes look fine to me.

@smayya337 what prevents this from pulling even the "All Good" status from the status page?

@dianalin2 fix CI, then you're good.

Laur04:
@dianalin2 you've got merge conflicts now. `dev` was updated recently, so rebase against it.

Laur04:
@dianalin2 don't worry about Codacy

Laur04 (changes requested):
LGTM - squash commits and I'll merge

Laur04:
@dianalin2 I didn't see it this time, so my bad, but in the future, please remove the old commit messages from the squashed commit message.

PR #1311, opened 2022-05-26
fix(events): integration with dashboard, special event display

alanzhu1:
## Proposed changes
- Shows the full date & time and events calendar icon if the event is shown on the dashboard. 
- Adds events in the "special" section to calendar (events that have the option checked for show attending, scheduled activity, announcement or show on dashboard). Not previously implemented due to styling issues. 

- If the event is shown on the dashboard, it should include which day the event is on instead of only the time (on the calendar view, it only shows time). 
- Icon is shown to match formatting on dashboard and hidden in calendar to save space.


Laur04 (approved):
LGTM

PR #1315, opened 2022-06-02
fix(events): show_all fix, mobile-friendly format

alanzhu1:
## Proposed changes
- The show_all button is currently visible for all users, but only works for admins. Takes out the show_all button for non-admins. 
- More mobile-friendly format for events; the text is currently overlapping.
- Import JS to make attend button work on roster view. 
- Small content changes: remove "View Roster" button from the view roster page, remove redundant "no events" text.

### DIFF
--- a/intranet/templates/events/event.html
+++ b/intranet/templates/events/event.html
@@ -123,11 +123,6 @@
         <div class="bottom-row show-attend">
             <span class="signup-status">
                 People attending: {{ event.attending.count }}
-
-                <a href="{% url 'event_roster' event.id %}" class="button small-button view-roster-button">
-                    View Roster
-                </a>
-
             </span>
 
             {% if request.user in event.attending.all %}
--- a/intranet/templates/events/home.html
+++ b/intranet/templates/events/home.html
@@ -73,7 +73,7 @@
                 </a>
             {% endif %}
 
-            {% if not show_all %}
+            {% if not show_all  and is_events_admin %}
                 <a href="?show_all=1" class="button">
                     Show All
                 </a>
@@ -99,7 +99,7 @@
             {% endif %}
 
             {% if not show_all and not classic %}
-                &nbsp;
+                &nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;
                 <a id="month-button" class="button" href="#">Month</a>
                 <a id="week-button" class="button" href="#">Week</a>
                 &nbsp;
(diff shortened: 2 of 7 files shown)


intranet/templates/events/roster.html
```diff
@@ -51,7 +52,23 @@ <h2>Public Roster</h2>
 
             {% if num_hidden_members > 0 %}
                 <br>
-                <p>{{ num_hidden_members }} {% if num_hidden_members == 1 %}user {% if not request.user.can_view_eighth %}(you) {% endif %}has{% else %}users {% if not request.user.can_view_eighth %}(including you) {% endif %}have{% endif %} not given permission to appear on this list.</p>
```
Laur04, line 54: Since this will all appear on one line on the page, it's best to leave it on one line now.

intranet/templates/events/event.html
```diff
@@ -123,11 +123,6 @@ <h3>
         <div class="bottom-row show-attend">
             <span class="signup-status">
                 People attending: {{ event.attending.count }}
-
-                <a href="{% url 'event_roster' event.id %}" class="button small-button view-roster-button">
-                    View Roster
-                </a>
-
```
Laur04, line 130: I'm not understanding your logic here. You still want people to be able to see who is coming.

intranet/templates/events/home.html
```diff
@@ -73,7 +73,7 @@ <h2>Events</h2>
                 </a>
```
Laur04, line 73: Another thing that I've noticed is that unapproved events don't show up on the top. I'd like them to, so that only approved events show in the calendar.

Laur04 (changes requested):

intranet/templates/events/event.html
alanzhu1: Lines 15-17 already include a link to the roster:
` <a href="{% url 'event' event.id %}" class="event-link">{{ event.title }}</a>`
The /events/event/{id} URL path also renders the roster view (in addition to /events/roster), but I can add back the button if you would prefer that.

intranet/templates/events/home.html
alanzhu1: Unapproved events should have their own dedicated section at the top of the page:
`{% for category in events %}
    {% if category.title == "Awaiting Approval" and category.events %}
        <h2 class="category">{{ category.title }}:</h2>
            {% for event in category.events %}  
                 …
` (lines 125-135)

Laur04:
Rebase against latest `dev` to resolve the linting issues and then obviously take a look at your tests.

PR #1353, opened 2022-08-05
style(search): change all alumni to graduate in search results

Rushilwiz:
## Proposed changes
The search results should display all graduates as graduates and not "Grade 1X"


shahsalonik (approved):
LGTM

PR #1359, opened 2022-08-10
fix(dev): running docker as sudo

alanzhu1:
## Proposed changes
- Makes Git trust the root Ion directory for all containers.

- The Docker daemon needs root privileges, requiring the use of `sudo` or adding the user to the `docker` group. 
- When running Docker as `sudo` and the directory containing Ion's source code is owned by another user, Git will not run because the directory is not trusted.


Laur04 (approved):
LGTM, @shahsalonik

PR #1366, opened 2022-08-17
refactor: focus searchbar on keydown and better tab nav

alanzhu0:
## Proposed changes
- Typing goes to the search bar on dashboard page
- Tab goes to search bar -> View Eighth Profile -> left-side menu on profile page

- Email from Dr. Gabor

### DIFF
--- a/intranet/static/js/dashboard/common.js
+++ b/intranet/static/js/dashboard/common.js
@@ -3,10 +3,19 @@ $(function() {
     $('.dash-warning').click(function (e) {
         $(this).toggleClass('collapsed');
     });
+
     $('.widget.extra-widgets-show').click(function() {
         $('body').addClass('show-extra-widgets');
     });
+
     $(".new-feature-close").click(function(e) {
         $(e.target).closest(".new-feature").hide("slow");
+    });
+
+    $("body").on("keydown", function(e) {
+        if(e.keyCode === 9) {
+            e.preventDefault();
+        }
+        $("#searchbox").focus();
     });
 });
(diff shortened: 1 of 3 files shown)


Rushilwiz:
hah! you beat me to this, I was just working on it when I got the email.

alanzhu1:
Haha, love the race! Maybe your implementation is more elegant? Mine is pretty crude.
We probably should've communicated too lol

Rushilwiz:
For sure. I think your implementation is workable, but I think this can be achieved without JS. You're looking for the `autofocus` and `tabindex` fields, I think, HTML already has a lot of great ARIA stuff that does exactly what you want.

https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/tabindex

So just

```html
<input autofocus tabindex=0 type="text" name="q" value="{% if search_query %}{{ search_query }}{% endif %}" placeholder="Search directory and more" aria-label="Search directory and more">
```

should work, and then just increment the tab index on the other elements instead of using JS.

alanzhu1:
Yeah, that was actually exactly my first approach but I ran into two issues: 
1. I didn't want the cursor to be already blinking in the search field upon load, just for aesthetics (not a big issue)
2. Whenever I used `tabindex`, any value other than 0 didn't work - positive values acted as if they were negative values. Maybe it was a problem on my end? It was pretty weird.
I definitely agree an HTML-only approach would be better though.

shahsalonik (approved):
LGTM

PR #1390, opened 2022-09-09
isort fixes

alanzhu0:
## Proposed changes
- isort recursive flag fix
- make isort happy in a few files (sorry I missed this before since linting wasn't getting to isort because of pylint)


shahsalonik (changes requested):
This looks good, but can you squash your commits? After that, I'll merge it :)

alanzhu0:
Oh so actually these are supposed to be two different commits since they deal with different parts of the codebase - is that okay?

shahsalonik:
Yeah, so even though they deal with two different things, you should still squash them.

alanzhu0:
Oh okay I'll do that - I'm not on my computer right now but I'll get it done by tomorrow!

shahsalonik:
Yup, no big deal. Whenever you have time is good!

shahsalonik (approved):

PR #1400, opened 2022-09-18
fix(eighth): fix activity detail not showing

alanzhu0:
## Proposed changes
- Fix issue with activity detail view not showing 
- Also fix only showing waitlist button to admins even if waitlist is enabled

- Bug made detail view and waitlist not work (commit: #1385)


shahsalonik (approved):
LGTM (tysm :))

PR #1402, opened 2022-09-18
test(eighth): fix tests failing between 11-11:59pm

alanzhu0:
## Proposed changes
- Change `test_signup` to use `timedelta` 

- Tests fail between 11:00 and 11:59 pm because adding 1 to `now.time().hour` would have hour as 24


shahsalonik (approved):
LGTM

PR #1448, opened 2022-11-02
style(cslapps): added client secret warning for oauth

Rushilwiz:
## Proposed changes
Add explicit help text showing that the client's secret is hashed and hide it after it's made.


Rushilwiz:
LGTM

PR #1459, opened 2022-11-08
fix(bus): mobile zindex, other fixes

alanzhu0:
Fix bus map overlaying search bar on mobile
Show bus status as "not arrived yet" if not arrived 5 min before end time
Remove unneeded polls stylesheet
Remove console.log()s


shahsalonik (approved):
LGTM

PR #1465, opened 2022-12-07
refactor(themes): show disable halloween theme on nav bar, fix piday 

alanzhu0:
refactor: display the disable Halloween theme button as part of the nav bar on mobile
fix: only show piday theme on auth page, not on every page

This PR depends on #1464 (please only merge after the snow one has been merged).


shahsalonik (approved):
LGTM

PR #1481, opened 2023-01-04
Remove @login_required decorator for reset_password_view

jwoglom:
## Proposed changes
- Removes the `@login_required` decorator for reset_password_view, because it blocks expired password resets
- Adds a `sensitive_post_parameters` decorator for the old and new password POST parameters, as is also done in index_view

When a user's account has an expired password and the user cannot perform a regular log-in, a return value of `KerberosAuthenticationResult.EXPIRED` is given from `KerberosAuthenticationBackend`. The login form returns the `RESET_PASSWORD` service user, but does not log in to it before redirecting to the reset_password route.

However, the reset password page is locked behind requiring authentication. Since `login()` wasn't called on the returned user from the authentication form, the user is then redirected back to the login page, preventing them from logging in with the expired password which needs a reset. This manifests as a returned URL of `/login?next=/reset_password%3Fexpired%3DTrue`, which has the effect of, possibly, preventing a still-technically-active Kerberos user account of a certain TJ alumnus from authenticating to Ion.

The reset_password view should actually be safe to be exposed without authentication, since in order for it to function in the first place it requires a successful authentication with Kerberos (see https://github.com/tjcsl/ion/blob/master/intranet/apps/auth/helpers.py#L34). Another potential solution would be to log in to Django with the returned `RESET_PASSWORD` user account, but block access to actually perform any tasks as that service account besides changing the desired user's password; however, that would be a riskier and more complicated change requiring more testing.

### DIFF
--- a/intranet/apps/auth/views.py
+++ b/intranet/apps/auth/views.py
@@ -275,7 +275,7 @@ def reauthentication_view(request):
     return render(request, "auth/reauth.html", context)
 
 
-@login_required
+@sensitive_post_parameters("old_password", "new_password", "new_password_confirm")
 def reset_password_view(request):
     context = {"password_match": True, "unable_to_set": False, "password_expired": request.GET.get("expired", "false").lower() == "true"}
     if request.method == "POST":


alanzhu0 (approved):
Thank you! I apologize for the very long delay in merging this as it seems to have escaped previous Sysadmins's attention. We did indeed receive many questions relating to the problem you mentioned in your comment. Thank you again for this and for starting this entire project.

PR #1512, opened 2023-02-17
fix(dev): fix docker dev env image

NotFish232:
## Proposed changes
- use the -I (--ignore-installed) flag when installing requirements and dev requirements.

Building image without --ignore-installed flag throws an error about an existing installed package.
This flag fixes it and allows the dev env to work normally.


shahsalonik (changes requested):
Can you run the scripts the tests are saying you need and then re-commit?

Issue #1411, opened 2022-09-30
Hide Polls icon when no polls are running


alanzhu0:
## Feature description
- Hide the Polls button on the left-side navbar when there are no currently open polls for non-admins. 

## Rationale for adding feature
- It's not very useful to have it displayed when 95% of the time there's nothing to see.

## Suggested implementation(s) (if applicable)
- Maybe look at how the Printing button is hidden when outside TJ network.

PR #1520, opened 2023-04-05
feat(polls): fix poll visibility

remtbkv:
## Proposed changes
- Add function "should_see_polls" for User model
- Use it in a template tag in nav.html 

The function should_see_polls returns true if there are any active polls visible to the user or if the user has admin permissions to polls in which case the polls icon should always be available. This function is used in a template tag in nav.html to only show the polls icon when necessary for the user.

### DIFF
--- a/intranet/apps/users/models.py
+++ b/intranet/apps/users/models.py
@@ -970,6 +970,15 @@ class User(AbstractBaseUser, PermissionsMixin):
         now = timezone.localtime()
         return Poll.objects.visible_to_user(self).filter(start_time__lt=now, end_time__gt=now).exclude(question__answer__user=self).exists()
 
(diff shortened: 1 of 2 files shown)


shahsalonik (changes requested):
Looks good overall, but could you switch the branch you're pushing to from `master` to `dev` (according to our [PR guidelines](https://github.com/tjcsl/ion/blob/master/CONTRIBUTING.md#pull-requests))? Thanks!

alanzhu0 (changes requested):
Your function does not work. You need to filter the `Poll` objects by time. 

Also, in accordance with the PR guidelines, please tag #1411 in your commit message to automatically close it upon merge.

shahsalonik (approved):
LGTM

PR #1522, opened 2023-04-10
fix(dev): make docker build work for arm

disco-io:
fix(dev): make docker build work for arm


disco-io:
Added libffi-dev to the dependencies list so the Dockerfile could support ARM architecture.

shahsalonik (approved):
LGTM!

PR #1529, opened 2023-04-27
Fix formtools

alanzhu0:
## Proposed changes
- Fix `django-formtools` breaking `/eighth` for teachers
- Rollback to v2.3 after bump in https://github.com/tjcsl/ion/commit/e732f3bb27148068297ce8d56b93178db7ce85a2


alanzhu0:
Note: the commit history is very messy because `dev` is out of sync with `master` - they need to be synced again and then the PR will show only 1 commit

shahsalonik (approved):

PR #1540, opened 2023-05-07
feat(polls): add randomization option

### LINKED ISSUE #1526: Polls: Randomize choice order for elections
## Feature description
- The Polls app is often used for conducting school elections. To improve fairness, the choices (candidates) should appear in a randomized order when voting.

For example, the order on one page load may be:
1. Alice
2. Bob
3. Charlie 

And on a subsequent load may be:
1. Bob
2. Alice
3. Charlie

- This should only be enabled when `Poll.is_election` is true. 

- There is already an Election question type that randomizes the choices. However, SGA has recently moved to rank choice voting. This feature should randomize *all* question types when `Poll.is_election` is true.

disco-io:
## Proposed changes
- Add a randomization checkmark feature to polls

- Ensures fairness of voting

Closes #1526


ksquaring:
LGTM! @alanzhu0 check

alanzhu0 (changes requested):
Good work! Could you make it so questions can be individually randomized as we discussed?

shahsalonik (changes requested):
LGTM! Could you add in Alan's changes? Thanks!

disco-io:
Yes! I will work on it.

PR #1545, opened 2023-05-12
Rank choice fixes

alanzhu0:
## Proposed changes
- Fix election results breaking when there is only one choice for a question
- Add top choice for other question types in election results view
- Make directions clearer for rank choice (handle edge case of only 1 choice and make it clear the max choices applies only to the current question)


shahsalonik (approved):
LGTM

PR #1580, opened 2023-08-19
feat(ion): all .format calls -> f-strings

NotFish232:
## Proposed changes
- .format in python has been changed to f-strings
- change all instances of .format to f-strings

- f-strings are a better, more concise, version of .format
-
(Please review all of the changes Alan 谢谢）

### DIFF
--- a/intranet/apps/users/models.py
+++ b/intranet/apps/users/models.py
@@ -320,7 +320,7 @@ class User(AbstractBaseUser, PermissionsMixin):
             The user's full name (first + " " + last).
 
         """
-        return "{} {}".format(self.first_name, self.last_name)
+        return f"{self.first_name} {self.last_name}"
 
     @property
     def full_name_nick(self) -> str:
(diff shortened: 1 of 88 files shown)


intranet/apps/users/models.py
```diff
@@ -1299,10 +1297,10 @@ def __setattr__(self, name, value):
     def __getattr__(self, name):
         if name == "number":
             return self._number if self.user.properties.attribute_is_visible("show_telephone") else None
-        raise AttributeError("{!r} object has no attribute {!r}".format(type(self).__name__, name))
+        raise AttributeError(f"{type(self).__name__!r} object has no attribute {name!r}")
```
alanzhu0, line 1300: Not needed with f-strings - use `repr()` instead.

intranet/apps/users/models.py
```diff
@@ -1364,7 +1362,7 @@ def __setattr__(self, name, value):
     def __getattr__(self, name):
         if name == "binary":
             return self._binary if self.user.properties.attribute_is_visible("show_pictures") else None
-        raise AttributeError("{!r} object has no attribute {!r}".format(type(self).__name__, name))
+        raise AttributeError(f"{type(self).__name__!r} object has no attribute {name!r}")
```
alanzhu0, line 1365: See repr() comment

alanzhu0 (changes requested):
A few notes: 
- Please see repr() comments and fix all places they occur - I did not catch them all.
- Please change commit message to be something like `refactor: all .format calls -> f-strings` as this is not a feature. "ion" should not be included in the parentheses as it is not an appropriate subsection of the codebase.

Otherwise, looks good!

PR #1585, opened 2023-09-02
feat(hoco): Refresh hoco scores

### LINKED ISSUE #1476: Unintended behavior of hoco scores on signage
Commit https://github.com/tjcsl/ion/commit/7b2527ec806f0141fe321ad65f86f32a586262ba added hoco scores to signage. Although this has not been seen in production (hoco is a long time away), the scores will only update on page load. Signage pages only load once a day, so scores will be inaccurate if they are updated during the day (which will happen during hoco). This can easily be fixed by adding a timeout to call a function to load scores every time interval, say 5 minutes, in https://github.com/tjcsl/ion/blob/master/intranet/static/js/hoco_scores.js.

Pinging @KrishnanS2006 on this

KrishnanS2006:
## Proposed changes
- Refresh hoco scores every 5 seconds
- Tweak score box animation slightly

 - Resolves #1476 
 - A 2-second delay for score boxes syncs their animation with the hoco ribbon

### DIFF
--- a/intranet/static/js/hoco_scores.js
+++ b/intranet/static/js/hoco_scores.js
@@ -1,13 +1,38 @@
 $(document).ready(function() {
-    $.get("https://homecoming.tjhsst.edu/api/", function(data) {
-        console.log(data.senior_total ? data.senior_total : 0);
-        $("#score-senior").text(data.senior_total ? data.senior_total : 0);
-        $("#score-sophomore").text(data.sophomore_total ? data.sophomore_total : 0);
-        $("#score-junior").text(data.junior_total ? data.junior_total : 0);
-        $("#score-freshman").text(data.freshman_total ? data.freshman_total : 0);
-        giveRibbons();
-        $("#hoco-scores").fadeIn();
-    });
+    let hocoScoresRef;  // stores a reference to the scores div when disconnected
+    let numDisconnected = 0;
+    let didConnect = false;  // whether we've successfully connected at least once
+    function loadScores() {
+        $.get("https://homecoming.tjhsst.edu/api/", function(data) {
+            didConnect = true;
+            if (hocoScoresRef) {  // on reconnect
+                $("#hoco-scores").replaceWith(hocoScoresRef).fadeIn();  // restore the backup
+                hocoScoresRef = null;
+            }
+            $("#score-senior").text(data.senior_total ? data.senior_total : 0);
+            $("#score-sophomore").text(data.sophomore_total ? data.sophomore_total : 0);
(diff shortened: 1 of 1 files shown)


intranet/static/js/hoco_scores.js
```diff
@@ -1,13 +1,18 @@
...
-        $("#score-senior").text(data.senior_total ? data.senior_total : 0);
-        $("#score-sophomore").text(data.sophomore_total ? data.sophomore_total : 0);
-        $("#score-junior").text(data.junior_total ? data.junior_total : 0);
-        $("#score-freshman").text(data.freshman_total ? data.freshman_total : 0);
-        giveRibbons();
-        $("#hoco-scores").fadeIn();
-    });
+    function loadScores() {
+        $.get("https://homecoming.tjhsst.edu/api/", function(data) {
+            console.log(data.senior_total ? data.senior_total : 0);
```
alanzhu0, line 4: Should be removed

intranet/static/js/hoco_scores.js
```diff
@@ -1,13 +1,18 @@
...
+            $("#score-senior").text(data.senior_total ? data.senior_total : 0);
+            $("#score-sophomore").text(data.sophomore_total ? data.sophomore_total : 0);
+            $("#score-junior").text(data.junior_total ? data.junior_total : 0);
+            $("#score-freshman").text(data.freshman_total ? data.freshman_total : 0);
+            removeRibbons();
+            giveRibbons();
+        });
```
alanzhu0, line 11: In case the network is slow or unavailable, this AJAX query may fail. If it fails more than some number of times in a row, say 3, then I think it should display a message that it was unable to query hoco scores to avoid misinforming people.

I would implement this with a counter and `jQuery.ajax()`'s error attribute.

alanzhu0 (changes requested):
See two comments. Please also fix commit format. Good work!

intranet/static/js/hoco_scores.js
KrishnanS2006: Done in c718222 - it just hides hoco scores if the API is unreachable, and unhides them when it's reachable again.

alanzhu0: Please implement described counter and also squash commits.

alanzhu0 (changes requested):

alanzhu0 (changes requested):
LGTM. Please squash your commits.

alanzhu0 (approved):
LGTM. CI is broken right now.

PR #1610, opened 2023-11-29
fix(printing): fix disabled printers bug

krishnans2006:
## Proposed changes
- Make sure a printer's name (referred to as description) is not overridden by other printers

Fixes a bug that affects #1603


alanzhu0 (approved):
LGTM

PR #1614, opened 2023-12-08
fix(misc): option to import staff photos

NotFish232:
## Proposed changes
- Update django command to be able to import staff photos
- Staff photo importing should be automated like with students


alanzhu0 (approved):
LGTM

PR #1615, opened 2023-12-09
fix(user): fix default photos for teachers

NotFish232:
profile pictures don't work for teachers without selecting a preferred image


alanzhu0 (changes requested):
Per internal discussion, a feature for teachers to select their preferred profile photo is needed. It would also be nice for them to have an option to upload a custom photo.

PR #1625, opened 2024-03-16
Refactor(users): use `enum.Enum` for `GRADE_NUMBERS`

JasonGrace2282:
## Proposed changes
- Use an enum for `GRADE_NUMBERS` instead of a 2D array

Python enums are much more flexible and readable then a 2D array. You can read a nice answer about their benefits [here](https://stackoverflow.com/a/37601645)

## Comments
It is recommended that [attributes of enums should be in all caps](https://docs.python.org/3.8/library/enum.html) but in order to make the PR easier to review (smaller diff) I chose to make the attributes of the enum lowercase.


NotFish232:
Hi Aarush. Thanks for the PR! I understand (and agree) with you on the benefits of enums, but in this case, I think the current implementation is better. 2D arrays mesh much better with how django normally wants field choices organized (like in the grade_number field on photos, where you convert the enum to a 2D array anyways). Since User is such a core model in ion, this change would be hard to test, and may lead to unexpected behavior in other apps, for not much actual benefit in code quality. Because of that, we can't accept this change and encourage you to work on other issues you find instead — issues tagged with “good first issue” are a great place to start.

JasonGrace2282:
Makes sense, I'll close this PR then!

PR #1626, opened 2024-03-19
fix(cslapps): update App.image_url

krishnans2006:
## Proposed changes
- Use `CharField` instead of `URLField`
- Lower `max_length`

- The validators used by `URLField` don't allow for updating the field in Django Admin, when in the form `/static/img/cslapps/app.ico`.
- The current `max_length` is unnecessarily large (and could cause errors with some database backends).


NotFish232 (approved):
LGTM

PR #1611, opened 2023-11-29
Migrate from setup.py to pyproject.toml

pefoley2:
Based on https://setuptools.pypa.io/en/latest/userguide

## Proposed changes

Migrate configuration from setup.py to pyproject.toml, but leave a stub setup.py to avoid breaking setup.py test and friends.

## Brief description of rationale

Setuptools is discouraging direct usage of setup.py https://setuptools.pypa.io/en/latest/userguide/quickstart.html#transitioning-from-setup-py-to-setup-cfg

PR #1648, opened 2024-03-29
Migrate from setup.py to pyproject.toml

krishnans2006:
This builds on #1611 by resolving the requested changes. The original PR description is copied below, for future reference.

---

Based on https://setuptools.pypa.io/en/latest/userguide

## Proposed changes

Migrate configuration from setup.py to pyproject.toml, but leave a stub setup.py to avoid breaking setup.py test and friends.

Setuptools is discouraging direct usage of setup.py https://setuptools.pypa.io/en/latest/userguide/quickstart.html#transitioning-from-setup-py-to-setup-cfg


Ion.egg-info/SOURCES.txt
```diff
@@ -875,7 +875,6 @@ intranet/middleware/url_slashes.py
 intranet/settings/.gitignore
 intranet/settings/__init__.py
 intranet/settings/ci_secret.py
-intranet/settings/secret.py
```
krishnans2006, line 878: A rollback to this change was requested in the original PR. However, it seems to me that this change is both necessary and proper, since `secret.py` is not in source control and doesn't need to be part of the egg. Further testing may be necessary.

NotFish232 (approved):
LGTM

PR #1659, opened 2024-04-07
Add `pre-commit` and use ruff

JasonGrace2282:
A similar change has been merged into Tin, see https://github.com/tjcsl/tin/pull/21

If you are reviewing this PR, I suggest looking at it commit-by-commit instead of looking at the diff all at once.

## Proposed changes
- Use `pre-commit` for linting/formatting
- Ideally maintainers will add `pre-commit ci` so that if the pre-commit fails, fixes will be pushed without having to run scripts.
  - Otherwise when creating a virtual environment, you can run `pre-commit install` and it should run pre-commit locally whenever you commit
- Change from `black`+`autopep8` to `ruff`
  - Faster: image
  - Drop in replacement for `black`: https://docs.astral.sh/ruff/formatter/#black-compatibility
  - Flake8 plugins included
  - Configurable in `pyproject.toml`

Running scripts is annoying lol

## Summary of Changes
* Add `.pre-commit-config.yaml`
  * Also adds some config for `pre-commit ci` if it is decided to be added
* Add `pre-commit` to dependencies
* Changes test pipeline to run `pre-commit run --all-files` instead of scripts for formatting/linting
* Deprecate the format scripts

### Note
PR #1673 should be merged before this to see if all checks pass.
I just tested locally, there are about ~11 files that are changed after merging it. Once it's merged I'll push a commit updating this PR with those 11 files.

## Progress
- [X] `scripts/format.sh`
- [X] `scripts/lint.sh`
- [x] `scripts/build_sources.sh`
  - See #1669  ?
- [x] `scripts/static_templates_format.sh`

### DIFF
--- a/intranet/apps/eighth/tests/admin/test_admin_groups.py
+++ b/intranet/apps/eighth/tests/admin/test_admin_groups.py
@@ -531,7 +531,7 @@ class EighthAdminGroupsTest(EighthAbstractTest):
         self.make_admin()
 
         # Create users
-        users = get_user_model().objects.get_or_create(username="users")[0]
+        userA = get_user_model().objects.get_or_create(username="userA")[0]
         userB = get_user_model().objects.get_or_create(username="userB")[0]
 
         # Create groups
@@ -539,7 +539,7 @@ class EighthAdminGroupsTest(EighthAbstractTest):
         group_not_empty = Group.objects.get_or_create(name="Not Empty Group")[0]
 
         # Add users to group
-        group_not_empty.user_set.add(users)
+        group_not_empty.user_set.add(userA)
         group_not_empty.user_set.add(userB)
 
         response = self.client.post(reverse("eighth_admin_delete_empty_groups_view"))
@@ -547,7 +547,7 @@ class EighthAdminGroupsTest(EighthAbstractTest):
         self.assertNotIn(group_empty, Group.objects.all())
 
         # Add user to other group
-        group_not_empty.user_set.remove(users)
+        group_not_empty.user_set.remove(userA)
         group_not_empty.user_set.remove(userB)
 
         response = self.client.post(reverse("eighth_admin_delete_empty_groups_view"))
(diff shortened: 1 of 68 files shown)


JasonGrace2282:
To reiterate what I said on matterless:
The current problem is (to the best of my knowledge) that `black` and `autopep8` interfere with each other. Black will format some files, then autopep8 will go in and fix those. This was fine in scripts where it simply merged both outputs (with autopep8 taking priority), but with `pre-commit` it counts a check as a fail if it modifies files. As such, both black and autopep8 will fail each run.
The possible solutions to this I can think of include:
* Remove `autopep8` and stick with `black`
* Remove `black` and only use `autopep8`
* Don't use either, and use something like `ruff`
  * I'm a little bit biased, but this is my preferred option
* Do something like:
```yaml
- repo: local
  hooks:
    id: format
    entry: ./scripts/format.sh && test -z "$(git status --porcelain=v1)"
```
which I'm not sure would work with `pre-commit ci`

JasonGrace2282:
I think this PR is _mostly_ ready to be reviewed, just need to find out what's causing the chaos with sorting imports.
Some benchmarks on my system:
* `time pre-commit run --all-files` takes 11.32 seconds (user)
* `time ruff check intranet/**/*.py` takes 0.45 seconds (user)
* `time ruff lint intranet/**/*.py` takes 0.04 seconds (user)

JasonGrace2282:
In order to get the tests to pass (specifically flake8) in #1673, I had to make some changes that don't agree with the ruff (or black for that matter!) in this PR. As such, the changes in that PR aren't 100% reflective of the changes needed here (but it's a pretty small and easy to review diff).

intranet/apps/eighth/tests/admin/test_admin_groups.py
NotFish232, line 1: This diff doesnt look right, make sure to look through all the changes before review :)
Some of the other changes seem to have the same issues as the last massive diff, i.e. the bad_strings from fcps

NotFish232 (changes requested):

intranet/apps/eighth/tests/admin/test_admin_groups.py
JasonGrace2282: Oops, forgot to check that! Should be fixed now (permanently, hopefully), thanks for taking the time to review :)

PR #1666, opened 2024-04-28
feat(printing): add rate limit to print job requests (closes tjcsl#662)

aarushtools:
## Proposed changes
- Add a rate limit of 10 POST requests per minute to ion printing
- Fix some linting errors from previous contributors in the printing/views.py file so checks can pass

Blocks excessive printing by users to prevent abuse or overload of the printers

### DIFF
--- a/intranet/apps/printing/views.py
+++ b/intranet/apps/printing/views.py
@@ -29,7 +29,38 @@ logger = logging.getLogger(__name__)
 
 
 class InvalidInputPrintingError(Exception):
-    """An error occurred while printing, but it was due to invalid input from the user and is not worthy of a ``CRITICAL`` log message."""
+    """An error occurred while printing, but it was due to invalid input from the user and is not worthy of a
+    ``CRITICAL`` log message."""
+
+
+class RatelimitCacheError(Exception):
+    """An error occurred while accessing the cache to rate limit a user"""
+
+
+class RatelimitExceededError(Exception):
+    """An error occurred because the user exceeded the printing rate limit"""
+
+
+def get_user_ratelimit_status(username: str) -> bool:
+    cache_key = f"printing_ratelimit:{username}"
+    value = cache.get(cache_key, None)
+    if value is None or value < settings.PRINT_RATELIMIT_FREQUENCY:
+        # User did not go over the rate limit
+        return False
+    elif value >= settings.PRINT_RATELIMIT_FREQUENCY:
+        return True
+    else:
+        raise RatelimitCacheError("An error occurred while trying to get your rate limit status")
+
+
+def set_user_ratelimit_status(username: str) -> None:
+    cache_key = f"printing_ratelimit:{username}"
+    value = cache.get(cache_key, None)
+    if value is None:
+        # Set the key to expire in the time specified by settings and indicate the user has requested once so far
+        cache.set(cache_key, 1, settings.PRINT_RATELIMIT_MINUTES * 60)
+    elif value >= 1:
+        cache.incr(cache_key)
 
 
 def get_printers() -> Dict[str, str]:
@@ -72,15 +103,15 @@ def get_printers() -> Dict[str, str]:
                     # Record the name of the printer so when we parse the rest of the
                     # extended description we know which printer it's referring to.
                     last_name = name
-            elif last_name is not None:
-                match = DESCRIPTION_LINE_RE.match(line)
-                if match is not None:
-                    # Pull out the description
-                    description = match.group(1)
-                    # And make sure we don't set an empty description
-                    if description:
-                        printers[last_name] = description
-                        last_name = None
+                elif last_name is not None:
+                    match = DESCRIPTION_LINE_RE.match(line)
+                    if match is not None:
+                        # Pull out the description
+                        description = match.group(1)
+                        # And make sure we don't set an empty description
+                        if description:
+                            printers[last_name] = description
+                            last_name = None
 
         cache.set(key, printers, timeout=settings.CACHE_AGE["printers_list"])
         return printers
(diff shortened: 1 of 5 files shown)


NotFish232 (changes requested):
Hi Aarush, thanks for the PR! Nice implementation with SImpleRateThrottle. I have a couple of suggestions.
1. A rate of 10/minute is probably too lenient, maybe try something more along the lines of a request every 5 minutes?
2. The current implementation ratelimits even if the request is not successful, i.e. the form is invalid. If a user tries to print and accidentally selects the wrong options (without an actual print job requested), they shouldn't have to wait for the ratelimit to expire.
Let us know if you need any help with this!

alanzhu0:
I think 2 requests / 5 minutes is a good ratelimit for successful POST requests.

I don't think turning the printing view into an API view is a good idea. It doesn't give a good error screen when the requests are throttled. Instead, the user should be presented with a good error screen, perhaps using the messages module we already have, and told they cannot print more than twice every 5 minutes. Perhaps you will find another package to accomplish this, but keep in mind only successful POST requests should be tracked. Otherwise, you may need to write your own backend and model for this.

aarushtools:
I tried to use django-ratelimit but decided that the package is just not meant to do what I was looking for and decided to use caching to make a sort of DIY ratelimit. I implemented the django cache system into two functions that I check for directly in the `print_job` function. The cache key holds an int that is incremented each time a request is made, and it expires in 300 seconds or 5 minutes.

alanzhu0 (changes requested):
This is good! Can you make the ratelimit a setting variable, instead of directly hardcoded? For example, 

```
# 2 print jobs every 5 minutes
PRINT_RATELIMIT_FREQUENCY = 2
PRINT_RATELIMIT_MINUTES = 5
```
 would control the ratelimit behavior.

aarushtools:
Done 👍 
I put those variables in the main settings file for everything

aarushtools:
Done 👍
I put those variables in the main settings file for everything

aarushtools:
Done 👍
I put those variables in the main settings file for everything

alanzhu0:
This is great! Last few comments:
- Can you write some text on the printing page alerting users to the ratelimit? Use the variables from the settings file.
- Can you make the ratelimit not apply to administrators (`admin_all`)? Still send a message along the lines of "you have reached the ratelimit of 2 req / 5 min, but since you are an administrator it will still go through." (for debugging purposes).
- Same thing for teachers - except don't send the message. So no ratelimit at all for teachers. While you're at it, maybe increase the number of pages teachers can print? 50 is probably a good value. That can go in settings too.

aarushtools:
Everything implemented except for point #2. I use `logger.error()` to send a message since `messages.error` requires `request` as a property, which I don't have access to in the `print_job` function. I could add request to the obj object or as a param for the print_job func, but I think that isn't really necessary (but I can still do it if you believe it's the best way)

alanzhu0:
That's fine. Good call. Almost ready to merge, I'm assuming you'll fix CI, last thing it looks like "WizardoWaldo" (your alt?) is included in the commit. I don't really care about that but if you do you might want to take that off.

aarushtools:
Yea, sorry about that I'm trying to fix it. Pushing using the PyCharm gui did that for some reason

aarushtools:
I'm not sure why the linting is failing. It seems like it has to do with modules I never edited?

intranet/templates/printing/print_form.html
```diff
@@ -23,6 +23,7 @@
     <b>The following restrictions apply:</b>
     <ul>
         <li>Print jobs can have a maximum of {{ DJANGO_SETTINGS.PRINTING_PAGES_LIMIT }} pages.</li>
```
alanzhu0, line 25: You need to change this to reflect the new teacher/student settings.

intranet/templates/printing/print_form.html
```diff
@@ -23,6 +23,7 @@
...
+        <li>You can send {{ DJANGO_SETTINGS.PRINT_RATELIMIT_FREQUENCY}} print jobs every {{ DJANGO_SETTINGS.PRINT_RATELIMIT_MINUTES }} minutes.</li>
```
alanzhu0, line 26: Small reword - You may send *up to* XX print jobs every XX minutes.

intranet/apps/printing/views.py
```diff
@@ -349,17 +380,44 @@ def print_job(obj: PrintJob, do_print: bool = True):
...
 
+        elif num_pages > settings.PRINTING_PAGES_LIMIT_TEACHERS and obj.user.is_teacher:
+            raise InvalidInputPrintingError(
+                f"This file contains {num_pages} pages. " f"You may only print up to {settings.PRINTING_PAGES_LIMIT_TEACHERS} pages using this tool."
+            )
+
+        if get_user_ratelimit_status(obj.user.username):
+            # Bypass rate limit for admins but still send error message for debugging purposes
+            if obj.user.has_admin_permission("printing"):
+                logger.error(
```
alanzhu0, line 408: logger.debug is more appropriate here

intranet/apps/printing/views.py
```diff
@@ -349,17 +380,44 @@ def print_job(obj: PrintJob, do_print: bool = True):
         if obj.page_range:
             if not range_count:
                 raise InvalidInputPrintingError("You specified an invalid page range.")
-            elif range_count > settings.PRINTING_PAGES_LIMIT:
+            elif range_count > settings.PRINTING_PAGES_LIMIT_STUDENTS and obj.user.is_student:
```
alanzhu0, line 383: There are user types other than students and teachers. Apply the teacher limit to teachers and counselors (and, while we're at it, if the user is an admin they can get the teacher limit too). Apply the student limit to everyone else. Same logic for ratelimiting. No ratelimit for teachers and counselors. Ratelimit for everyone else (except admins).

alanzhu0 (changes requested):

intranet/apps/printing/views.py
alanzhu0: Actually, Justin reminded me `is_teacher` checks for teacher or counselor. So that's ok. But you will still need to cover other user types.

aarushtools: I just made the first if statement check for printing admin perms or if they are a teacher, and anyone else gets the student limit because of the order the elifs appear in

intranet/templates/printing/print_form.html
```diff
@@ -22,7 +22,14 @@
     <br>
     <b>The following restrictions apply:</b>
     <ul>
-        <li>Print jobs can have a maximum of {{ DJANGO_SETTINGS.PRINTING_PAGES_LIMIT }} pages.</li>
+        {% if user.is_teacher or user.has_print_permission %}
+        <li>Print jobs can have a maximum of {{ DJANGO_SETTINGS.PRINTING_PAGES_LIMIT_TEACHERS }} pages.</li>
+        {% else %}
+        <li>Print jobs can have a maximum of {{ DJANGO_SETTINGS.PRINTING_PAGES_LIMIT_STUDENTS }} pages.</li>
+        {% endif %}
+        {% if user.is_student %}
```
alanzhu0, line 30: Small nitpick. Following convention this should be `not user.is_teacher and not user.has_print_permission`

alanzhu0 (changes requested):
Very very close. Sorry for all the comments about small details. 

Can you do a favor for me? `user.has_print_permission` is a weird property. All the other admin ones are named like `user.is_printing_admin`. Could you global replace `has_print_permission` with `is_printing_admin`? This isn't strictly related to your PR but it'd be a good change long-term for Ion.

aarushtools:
> Very very close. Sorry for all the comments about small details.

No worries, hopefully it's all good now 👍

intranet/apps/printing/views.py
```diff
@@ -349,17 +380,45 @@ def print_job(obj: PrintJob, do_print: bool = True):
...
+            if obj.user.is_printing_admin:
+                logger.debug(
+                    """Administrator %s passed the rate limit of %s print jobs every %s minutes, but since they are an
+                    administrator the request will still go through.""",
+                    obj.user.username,
+                    settings.PRINT_RATELIMIT_FREQUENCY,
+                    settings.PRINT_RATELIMIT_MINUTES,
+                )
+            # If user needs to be rate limited
+            elif obj.user.is_student:  # Don't rate limit teachers
```
alanzhu0, line 417: This goes back to what we were discussing earlier - **if** printing admin or teacher, bypass, **else** apply ratelimit logic. Not just students.

alanzhu0 (changes requested):

alanzhu0:
Logic looks good now, just resolve the merge conflicts

alanzhu0 (approved):

PR #1668, opened 2024-05-02
Fix depends_on attributes in docker compose

krishnans2006:
## Proposed changes
- Use the service name (`django`) instead of the container name (`intranet_django`) in `depends_on` attributes

Before:
```
service "celery" depends on undefined service "intranet_django": invalid compose project
```

After:
```
[+] Building 0.3s (11/11) FINISHED
```

This bug was introduced in f041203106924e162483a570b8d5c450026253be (PR #1658).


NotFish232 (approved):
LGTM! As a side note, its weird that this worked for me on my computer but not for others.

PR #1674, opened 2024-05-03
docs: fix log in instructions

aarushtools:
## Proposed changes
- Create a `USERNAMES.md` file that has a list of usernames pulled from the django-admin panel that you can log into. Contains info about superuser status and more
- Fix the setup.md instructions. Clarifies that you must use a pre-defined username, such as the ones in `USERNAMES.md` to log in.

- Currently, [setup.md](https://github.com/tjcsl/ion/blob/master/SETUP.md) tells you to log in using **your Ion username** and the password `notfish`. However, this does not work. The suggested changes will clarify the instructions to indicate which usernames you can log in with


NotFish232 (changes requested):
Thanks for this documentation change, the old version definitely needed an update for how to log in. In its current form, the dev env generates random data based on fake names (you can check out the scripts in config/scripts) so the list of users that you can sign in with varies between each dev env. The only ones that are constant, are admin and the 5 students / admins created for each grade level. Other than that this change looks good!

aarushtools:
I think everything should be fixed now

alanzhu0:
Looks good, can you squash your commits?

aarushtools:
All commits squashed 👍

aarushtools:
> Looks good, can you squash your commits?

All commits squashed 👍

NotFish232 (approved):
LGTM

PR #1676, opened 2024-05-04
fix(scripts): fix scope regex to match convention commit specification

JasonGrace2282:
The conventional commit grammar allows the scope to be any utf-8 characters, except for newlines and parenthesis (see [here](https://github.com/conventional-commits/parser#the-grammar)).
Currently, we only accept the lowercase English alphabet.


NotFish232 (approved):
LGTM

PR #1678, opened 2024-05-06
fix: enforce announcement visibility restrictions (closes tjcsl#1677)

aarushtools:
## Proposed changes
- Stop restricted announcements from being found via search and seen via announcement permalink if it is not supposed to be seen by the user
   
- Announcements that are not visible to a user's group on the home page are still visible via permalink and search. The suggested changes will make the visibility restrictions consistent across the board so users do not see an announcement they are restricted from

### DIFF
--- a/intranet/apps/announcements/views.py
+++ b/intranet/apps/announcements/views.py
@@ -288,6 +288,10 @@ def view_announcement_view(request, announcement_id):
     """
     announcement = get_object_or_404(Announcement, id=announcement_id)
 
+    if not announcement.is_visible(request.user):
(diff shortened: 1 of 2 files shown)


aarushtools:
Should I add fix(announcements) to the commit or just keep it at fix because I modified the search app too?

alanzhu0:
fix(announcements) is fine.
Don't manually check group membership - use this instead: https://github.com/aarushtools/ion/blob/16092e04d1656099cc349de2c559797573ec3ca0/intranet/apps/announcements/models.py#L16

alanzhu0 (approved):
Good work!

Issue #1663, opened 2024-04-15
CSL Apps Toolbar links slow


alanzhu0:
Opening CSL apps from the "Apps" toolbar in the top right is slow when the app is associated with an OAuth application and the user is accessing the app for the first time in 24 hours - Ion needs to update the latest access token.

Look into ways to optimize this behavior and reduce load time - currently it can be over 10 seconds. I think it is probably the below SQL query:
https://github.com/tjcsl/ion/blob/dev/intranet/apps/cslapps/views.py#L29

devsrivv:
Whenever you get the chance can you assign this issue to me? Thanks

NotFish232:
Done!

PR #1682, opened 2024-05-11
fix: add values_list("token, "expires") to reduce csl apps load time (closes Issue #1663)

devsrivv:
## Proposed changes
- Add `values_list("token, "expires")` in between `.order_by("-expires")` and `.first()`

Instead of retrieving all of the fields from the AccessToken object (which leads to slower load times), only retrieve necessary fields to filter the most recent Token. In turn, this will improve the load time of CSL apps.

#1663

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -32,14 +32,14 @@
                 },
                 {
                     "name": "Set up Python ${{ matrix.python-version }}",
-                    "uses": "actions/setup-python@v2",
+                    "uses": "actions/setup-python@v5",
                     "with": {
                         "python-version": "${{ matrix.python-version }}"
                     }
                 },
                 {
                     "name": "Set up pip cache",
-                    "uses": "actions/cache@v2",
+                    "uses": "actions/cache@v4",
                     "with": {
                         "path": "~/.cache/pip",
                         "key": "pip-${{ matrix.python-version }}"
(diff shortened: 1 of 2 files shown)


NotFish232 (changes requested):
Hey Dev, thanks for the PR! Could you please provide some numbers indicating the speed up this change would have in loading csl apps? Thanks!

devsrivv:
Hey Justin, I can't figure out a way to practically measure the speed of this change in my dev environment as the links lead to CSL apps of the actual ion (signed in to my account instead of the test account). However, I tested the change in my dev environment and I can confirm that it functions.

Originally, the query filtered through all 10 fields `id, user, source_refresh_token, token, id_token, application, expires, scope, created, updated` from `oauth2_provider.models.AccessToken`

I made it so that it only queries the `token` and `expires` fields along with the necessary `user` and `application` fields so that it saves time (reducing the size from 10 fields to 4 fields), which theoretically saves 60% more time.

Does the logic behind my solution make sense? If not, is there any way to check the issue in my dev environment? Thanks

alanzhu0:
I'm not sure `values_list` saves time here, which is why a quantitative measurement is needed. It seems to me what's taking a long time is filtering through the `AccessToken` model, not accessing the fields, but maybe your solution works. You could create your own OAuth application to test the timing. Have you done the Django blog assignment? That would work.

PR #1687, opened 2024-05-18
chore: changes to let lint checks pass

aarushtools:
## Proposed changes
- Declare variables as `None` in the problematic functions at the start, and assume they will be set later on so pylint doesn't complain

This makes linting checks pass (I think)


alanzhu0 (changes requested):
Your fix works, but the comments next to the fix lines aren't entirely accurate. We don't know it will be set later on. It will just be set if needed.

aarushtools:
I just removed the comments instead, if that's fine?

alanzhu0 (approved):

PR #1690, opened 2024-05-19
fix(bus): altering dimensions of bus.html tjcsl

DhanvinG:
## Proposed changes

Changed: 
` .bus-announcement-container {
        position: absolute;
        top: 55%;
        right: 6%;
        width: 50%;
        padding: 20px !important;
        font-size: 30px;`
To:

`        .bus-announcement-container {
            position: absolute;
            top: 85%;
            right: 3%;
            width: 25%;
            padding: 20px !important;
            font-size: 20px;`

Changed: 

`    #morning .bus-announcement-container {
        top: 50%;
        right: 3%;
        width: 27%;
    }`

To:

`        #morning .bus-announcement-container {
            top: 80%;
            right: 3%;
            width: 25%;
        }`

I reduced the font size, made the assignment buttons smaller, and moved the location of the assignment buttons from the center of the page toward the bottom so it wouldn't interfere with the bus images to the left.

Using the singular file -  just to show the changes in the two announcement buttons, but on the actual signage it will be different, here is the difference between before and after the edits:

Before:

Originalimage

After:

Updatedimage

### DIFF
--- a/intranet/templates/signage/pages/bus.html
+++ b/intranet/templates/signage/pages/bus.html
@@ -216,12 +216,12 @@
 
     .bus-announcement-container {
         position: absolute;
-        top: 55%;
-        right: 6%;
-        width: 50%;
+        top: 85%;
+        right: 3%;
+        width: 25%;
         padding: 20px !important;
 
-        font-size: 30px;
+        font-size: 20px;
         text-align: center;
         padding: 20px;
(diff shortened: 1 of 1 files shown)


alanzhu0:
Those screenshots don't look like what the bus page should look like.

PR #1691, opened 2024-05-20
fix(eighth): display correct 8th period block (closes #1660)

### LINKED ISSUE #1660: UI error when taking attendance for a past eighth period
## Steps to reproduce
1. Try to take attendance for an eighth period that has happened on a previous day

## Expected behavior
Date shows up properly, with the input field date matching the header date

## Actual behavior
Input field and actual date differ
Screenshot from 2024-04-09 08-42-38

hstievat:
## Proposed changes
- Add show_all_blocks GET parameter to attendance form
- Start 8th block selection list at the correct starting date, using this parameter

A show_all_blocks parameter is consistent with the 8th period block selection form. By adding the option to begin blocks at the beginning of the school year, blocks from the past will appear in the list when selected in the form from before. This eliminates the UI bug that occurred when selecting an 8th period in the past.

### DIFF
--- a/intranet/apps/eighth/views/attendance.py
+++ b/intranet/apps/eighth/views/attendance.py
@@ -166,7 +166,10 @@ class EighthAttendanceSelectScheduledActivityWizard(SessionWizardView):
         else:
             url_name = "eighth_take_attendance"
 
-        return redirect(url_name, scheduled_activity_id=scheduled_activity.id)
+        response = redirect(url_name, scheduled_activity_id=scheduled_activity.id)
+        if self.request.GET.get("show_all_blocks", "") == "1":
+            response["Location"] += "?show_all_blocks=1"
+        return response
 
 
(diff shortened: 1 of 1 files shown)


NotFish232 (changes requested):
Nice work on this feature, the functionality looks correct! Could you make it check that the value of the show_all_blocks key is explicitly equal to 1? Also I am not sure why you have the alternative option if block is given as a get parameter, you should remove that part unless you have a specific reason.

hstievat:
I copied that line of code directly (line 77 in the current version), to keep it as consistent as possible. Is there a reason it should be different? It should be no trouble to fix!

NotFish232:
There doesn't seem to be a reason to support using block as a parameter, and I would air on the side of having code that has less unintentional side effects. It should be a quick change.

NotFish232 (changes requested):
Change looks great! Can you please squash your commits and then it should be good to merge.

NotFish232 (approved):
Looks good, thanks for the PR!

PR #1696, opened 2024-06-10
fix(eighth): correct stats dates

NotFish232:
## Proposed changes
- Make stats correct based on actual eighth period dates

- current stats are wrong
- See activity 3261
- dates should be given within the data object in Chart.js, otherwise it goes chronologically. If num data entries is less than labels then it won't fill properly, although even if they are the same length the content is still wrong.


alanzhu0 (approved):

NotFish232:
Alan next time please wait for CI to finish first. Thanks.

PR #1697, opened 2024-06-12
chore: update typehints via `__future__.annotations`

JasonGrace2282:
Yay one less annoying thing when at some point dependencies are bumped

### DIFF
--- a/intranet/apps/users/models.py
+++ b/intranet/apps/users/models.py
@@ -1,8 +1,9 @@
 # pylint: disable=too-many-lines; Allow more than 1000 lines
+from __future__ import annotations
+
 import logging
 from base64 import b64encode
 from datetime import timedelta
-from typing import Collection, Dict, Optional, Union
 
 from dateutil.relativedelta import relativedelta
 from django.conf import settings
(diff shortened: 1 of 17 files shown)


alanzhu0:
@NotFish232 review

intranet/settings/__init__.py
```diff
@@ -202,7 +204,7 @@
 EMAIL_FROM = "ion-noreply@tjhsst.edu"
 
 # Use PostgreSQL database
-DATABASES = {"default": {"ENGINE": "django_prometheus.db.backends.postgresql", "CONN_MAX_AGE": 30}}  # type: Dict[str,Dict[str,Any]]
+DATABASES: dict[str, dict[str, str | int]] = {"default": {"ENGINE": "django_prometheus.db.backends.postgresql", "CONN_MAX_AGE": 30}}
```
NotFish232, line 207: like line 560 the type of this should probably be `dict[str, dict[str, Any]]` in case of new keys / values

intranet/utils/html.py
```diff
@@ -1,9 +1,14 @@
...
+
 import urllib.parse
-from typing import Mapping, Optional, Tuple, Union
+from typing import TYPE_CHECKING, MutableMapping
 
 import bleach
 from bleach.css_sanitizer import CSSSanitizer
 
+if TYPE_CHECKING:
+    Attr = MutableMapping[str | tuple[str | None], str]
```
NotFish232, line 10: defining a type for this seems unnecessary, I would just fix the function types without this

NotFish232 (changes requested):

intranet/apps/users/models.py
```diff
@@ -42,14 +44,14 @@ class UserManager(DjangoUserManager):
 
     """
 
-    def user_with_student_id(self, student_id: Union[int, str]) -> Optional["User"]:
+    def user_with_student_id(self, student_id: int | str) -> "User | None":
```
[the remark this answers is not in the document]
JasonGrace2282: Fixed, I (incorrectly) assumed the quotes were there for a reason.

intranet/utils/html.py
JasonGrace2282: You asked for it, let me know how you feel after looking at the typehints now ;)

PR #1699, opened 2024-07-01
fix(logs): rows are bright even on dark mode (closes tjcsl#1642)

aarushtools:
## Proposed changes
- add dark/logs.scss
- use css to style even table rows instead of js 

- fixes issue where even rows were white instead of a darker color on dark mode

before (from the issue page):
image

after: 
image


aarushtools:
Should I collectstatic and add those files into my commit or is it fine?

alanzhu0 (approved):

PR #1702, opened 2024-07-05
fix(signage): fix text containers overlapping bus spots in portrait mode (closes tjcsl#1680)

aarushtools:
## Proposed changes
- Change `padding `and `left` dynamically if `sign.landscape` is not true

Prevents the containers from overlapping with the bus spots

before:
before
after:
after

### DIFF
--- a/intranet/templates/signage/pages/bus.html
+++ b/intranet/templates/signage/pages/bus.html
@@ -123,6 +123,11 @@
         padding: 6px 10px;
         margin-bottom: 6px;
         margin-right: 6px;
+
+        {% if not sign.landscape %}
+        right: 1%;
+        padding: 12px !important;
+        {% endif %}
     }
 
     .logo {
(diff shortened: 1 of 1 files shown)


alanzhu0:
Looks good - how did you get choose the width of your screen? We will have to do some testing unless you found a way to match the signage displays' aspect ratio.

aarushtools:
> Looks good - how did you get choose the width of your screen? We will have to do some testing unless you found a way to match the signage displays' aspect ratio.

I put my screen orientation into portrait via my display settings and my aspect ratio is 16:10 (I guess 10:16 in portrait). Now that I look at the original issue image again, it actually does seem like the text containers are going further than what I was able to reproduce in my env. 

I just tested ratios using chrome dev tools, and it seems the max ratio it can go before getting smushed up next to the spots like this:


is at around `11:16`. It should be easy to make adjustments so it supports a higher ratio, just let me know what ratio I should aim for

alanzhu0 (approved):
I'm not sure what the actual ratio is, but this is good for now. Let's see what it looks like when school starts.

PR #1714, opened 2024-08-25
feat!: modernize documentation setup

JasonGrace2282:
This commit automatically documents Ion apps (removing the need for a script), and fixes many of the formatting errors in the current docs. It also updates the docs for the formatting/linting CI, which was updated in #1659

Note that Ion still uses Sphinx, but some scripts may need to be updated (I updated all of the ones I noticed).

I just had this branch lying around, so I figured I would clean it up and see what the thoughts were.

> [!NOTE]
> I just stole the Ion logo and changed some colors/fonts to make the logo. If anyone has a better logo, feel free to suggest it.


JasonGrace2282:
Currently, it does not document _every_ file in intranet (I'm still undecided if that's useful or if it will unnecessarily increase build times). Doing so is not a big change (a quick script should do it), so I'm open to thoughts about whether it should do that.

NotFish232 (approved):
LGTM

PR #1716, opened 2024-09-08
feat(meta): add ios pwa loading screen

aarushtools:
## Proposed changes
- add a loading screen to ios when opening as a pwa

- makes it look more native since there won't be a white screen for a few seconds when opening

[Video of what it looks like](https://streamable.com/mhb0gc)


krishnans2006:
Looks amazing, great work! Just a quick question (not an official review): where'd you get that list of required resolutions? Feels slightly excessive to me...

alanzhu0:
Looks good! Small thing - can you make the ion logo background color match the blue overall loading screen background?

aarushtools:
> Looks amazing, great work! Just a quick question (not an official review): where'd you get that list of required resolutions? Feels slightly excessive to me...

Yeah it is but unfortunately you have to put a link tag for each device on the [apple spec](https://developer.apple.com/design/human-interface-guidelines/layout#Specifications). I used [pwa-asset-generator](https://github.com/elegantapp/pwa-asset-generator) and it auto generates most of this. Also, all the tags are abstracted away in a separate file (and images in a separate folder) so 1-2 more lines shouldn't really make a difference if we want to match 100% of the apple spec

> Looks good! Small thing - can you make the ion logo background color match the blue overall loading screen background?

Sorry I didn't notice, it should be fixed now 👍

alanzhu0 (approved):

PR #1722, opened 2024-09-21
fix: remove temp fix for importlib_metadata type error

shrysjain:
The TypeError mentioned in [1718](https://github.com/tjcsl/ion/pull/1718) has been resolved with `zipp` v3.20.2. This PR removes our temporary fix.


shrysjain:
Nevermind I don't think this applies to us, keep the fix for now.

PR #1723, opened 2024-09-25
Implement individual students stickying

### LINKED ISSUE #1719: Individual User Sticky to Activity
## Feature description
- Allow individual students to be stickied to an activity. Currently, only entire activities or scheduled instances of an activity can be marked as sticky. This feature would allow 8PO to specify individual students that cannot unsignup from an activity, while all other students can.

- Requested by Math/CS Dept for tutoring sessions.

- Currently, the sticky logic is binded only to an `EighthScheduledActivity`. There could be a field added to `EighthScheduledActivity` called `sticky_students` or similar. `EighthSignup` may need to have a sticky attribute. I haven't looked at the code yet so these attributes might already be in place.

https://ion.tjhsst.edu/eighth/admin/scheduling/schedule?activity=925
The schedule activity screen would be a good place to add this:
image
There could be a dropdown field added for stickied students (similar to the sponsor field).

JasonGrace2282:
## Proposed changes
Allow stickying individual students to a scheduled 8th period.

## Brief description of change
Closes #1719 
* Add a `ManyToManyField` to `User` called `sticky_students`.
* There is also a new method called `EighthScheduledActivity.is_user_stickied(user)` which checks if an individual user is stickied, replacing the old `get_sticky` method.

You can edit the stickied students at `/eighth/admin/scheduling/schedule?activity=<some_activity_pk>`


alanzhu0 (changes requested):
Great work - it looks like the core feature is working.

I don't know if you've noticed this, maybe it only occurs when you have a fairly large scheduling page (lots of instances), but the page load time is very slow. Run the dev autogenerate data scripts if you haven't. I'm getting 15s page load vs. 2s on production, and my dev is usually faster than prod. I assume it's due to populating all the fields with all those users. So we have to optimize this somehow. Maybe use JS and only populate the field when the user clicks on it?

And a few feature suggestions:
- Can you distinguish between activity-sticky activities and user-sticky activities to the student? Right now they look the same to the student. Probably make the badge say "User Sticky" with an explanation tooltip instead of just "Sticky."
- What do you think about sending an email to the user when they get user-sticked into an activity?

JasonGrace2282:
> Maybe use JS and only populate the field when the user clicks on it?

I just pushed a commit that populates it as soon as the page has loaded. Could you check if that's fast enough for you?

As for sending an email, I'm not sure how much I would like that as a user - but I could also see it being really helpful for some people. As such, I'm -0 on the idea: if you really think it's useful/important, I can implement it.

alanzhu0 (changes requested):
Ok, this is great, it loads very quickly now. But it doesn't seem to work anymore? When I put a user in and save it doesn't actually do anything, and the field gets wiped? Is it on my end?

alanzhu0:
Did you change the name to 'user sticky'? Also yes - email would be helpful if you can add that.

JasonGrace2282:
I haven't implemented email yet (if I get a chance later, I can do that), but everything else should be done, namely:
- Fixing the bug with ajax
- Changing the label to individual sticky

intranet/apps/eighth/views/admin/scheduling.py
```diff
@@ -147,7 +147,11 @@ def schedule_activity_view(request):
                             messages.error(request, f"Did not unschedule {name} because there is {count} student signed up.")
                         else:
                             messages.error(request, f"Did not unschedule {name} because there are {count} students signed up.")
-                    instance.save()
+
+                    if instance:
+                        instance.sticky_students.set(form.cleaned_data["sticky_students"])
+                        instance.members.add(*form.cleaned_data["sticky_students"])
```
alanzhu0, line 153: You also need to remove the user's current `EighthSignup` using `remove_signup` and create a new `EighthSignup` for the sticky activity. Just adding them to members causes bugs. It needs to be done the proper signup way (see the signup view for the proper way).

alanzhu0 (changes requested):
Looking good and almost there

intranet/apps/eighth/views/admin/scheduling.py
JasonGrace2282: In my latest commit [d726662](https://github.com/tjcsl/ion/pull/1723/commits/d726662fc6b13726af0b9c8e426365d226f613c5) I used `EighthSignup.objects.create_signup` in addition to adding the user to `self.members`. Is there anything else I should be aware of?

intranet/apps/eighth/models.py
```diff
@@ -1097,10 +1115,50 @@ def notify_waitlist(self, waitlists: Iterable["EighthWaitlist"]):
...
+        self.members.add(*users)
+        # note: this will send separate emails to each student for each activity they are stickied in
+        new_stickied_students = (user for user in users if user not in old_sticky_students)
+        unstickied_students = (user for user in old_sticky_students if user not in users)
+        email_send_task.delay(
+            "eighth/emails/students_stickied.txt",
+            "eighth/emails/students_stickied.html",
+            data={"activity": self},
+            subject="You have been stickied into an activity",
+            emails=[user.notification_email for user in new_stickied_students],
```
alanzhu0, line 1146: Not a big deal, but why not do it all in one list comprehension...

intranet/templates/eighth/emails/students_stickied.html
```diff
@@ -0,0 +1,16 @@
...
+  <body>
+    {% with description="Eighth Signup Changes" %}
+        {% include "email_metadata.html" %}
+    {% endwith %}
+
+    <p>
+      You have been stickied into {{ activity.activity.name }} on {{ activity.date }} ({{ activity.block_letter }}).
+
+      If there has been a mistake with this change, please contact the 8th period office.
+      Have a nice day!
```
alanzhu0, line 11: I dont think you need this lol

intranet/apps/eighth/models.py
```diff
@@ -1097,10 +1115,50 @@ def notify_waitlist(self, waitlists: Iterable["EighthWaitlist"]):
                 [waitlist.user.primary_email_address],
             )
 
+    def set_sticky_students(self, users: "Sequence[AbstractBaseUser]") -> None:
```
alanzhu0, line 1118: yes good

intranet/apps/eighth/models.py
```diff
@@ -1097,10 +1115,50 @@ def notify_waitlist(self, waitlists: Iterable["EighthWaitlist"]):
...
+                signup.remove_signup(user, force=True)
+            EighthSignup.objects.create_signup(user, scheduled_activity=self)
+
+        old_sticky_students = self.sticky_students.all()
+        self.sticky_students.set(users)
+        self.members.add(*users)
+        # note: this will send separate emails to each student for each activity they are stickied in
+        new_stickied_students = (user for user in users if user not in old_sticky_students)
+        unstickied_students = (user for user in old_sticky_students if user not in users)
+        email_send_task.delay(
```
alanzhu0, line 1141: i think this works

alanzhu0 (changes requested):

intranet/templates/eighth/emails/students_stickied.html
JasonGrace2282: It never hurts to tell people to have a nice day :)

PR #1733, opened 2024-10-23
feat: implement club announcement restrictions

JasonGrace2282:
This makes it so that announcements that are created by a restricted club are only shown to people who are allowed to access the restricted club.


JasonGrace2282:
> LGTM, just to check though did you run lint/build/test on local?

CI is working again, and it passes all 3 (with the exception of the known problem about gathering the csl status).

PR #1737, opened 2024-10-30
feat: fix club announcements for 15+ announcements

reverileo:
## Proposed changes
- Handle the case for 15+ announcements

Previously, the counter wouldn't go down as the user read the announcements. Now, it automatically fetches the next announcement as the user reads announcements off.


reverileo:
Closing this in favor of a new approach. See PR #1760.

PR #1740, opened 2024-12-11
fix(misc): import teacher photos should only find staff

NotFish232:
## Proposed changes
- importing teacher photos should only consider staff users

- It would be awfully unfortunate if a teacher and a student both had the same names and the wrong image was loaded


alanzhu0:
Yes that would be AWFULLY unfortunate

PR #1754, opened 2025-03-31
feat: add betterstatus embed

aarushtools:
## Proposed changes
- remove the old method of scraping the TJCSL status page
- switch to a betterstatus embed in the header

- cleaner and doesn't require manual scraping
- manual scraping doesn't work right now anyway


aarushtools:
What it looks like
Screenshot 2025-04-14 222600
Screenshot 2025-04-14 215146

Responsiveness test
https://github.com/user-attachments/assets/aac01ba4-d0b9-4c8f-b22f-ebe0ec9b5bd6

aarushtools:
> This looks nice! Are you also going to add the automatic pulling of status updates from the status page, or is that going to be a separate PR (or just dropping it entirely)?

I was thinking the way we make manual announcements for outages is best (because you can decide what to say rather than auto generated updates) and the user can click the new status icon when it's down to see the auto status updates if they wanted. If it's needed though I could do it in a future PR

PR #1757, opened 2025-04-15
fix(announcements): missing migration

aarushtools:
#1733 was missing a migration


JasonGrace2282:
> thanks for catching this (smh @JasonGrace2282 )

In my defense, it makes no sense that changing a help text requires a new migration.

PR #1766, opened 2025-04-21
feat: lock out user after number of attempts

amcsz:
## Proposed changes
- Lock user out for LOCK_OUT_MINUTES after MAX_LOGIN_FAILURES attempts. Both are specified in django settings
- The user will see a message saying they need to wait X minutes.
- After the time, the user can log in again, otherwise they cannot.
- Data for lockout is stored in session, meaning it is localized to one device only

- Prevents spammers/bots from spamming the system


amcsz:
ah, ok! I will close this pr and work on those changes. I still haven't fully grasped my head around PAM but I think I'm getting more familiar. thank you!

PR #1767, opened 2025-04-21
feat(auth): tell user if they are locked out

### LINKED ISSUE #1749: Tell users if their account has been locked due to too many login attempts
## Feature description
If a user authenticates with the wrong password too many times, they will be locked out by IPA. This should be returned in the PAM error message. Ion should let the user know they have been locked out and that they will have to wait X minutes before logging in again. 

- Message on login screen

amcsz:
Closes #1749

## Proposed changes
- New state added to PamAuthenticationResult to indicate that the user is locked out.
- The authenticate method will update the session data, prompting the index view to show a message to the user, telling them they are locked out. The message goes away once the user reloads the page and will only appear when the user attempts to login while locked out.
- The authenticate method is called in the post_view and the request object is passed through so it is able to access the session data.

The user will know that they are locked out instead of just knowing that the login did not work.

### DIFF
--- a/config/docker/Dockerfile
+++ b/config/docker/Dockerfile
@@ -1,36 +1,18 @@
-FROM python:3.8.19
+FROM python:3.8.19-alpine3.18
 
-ENV TZ=America/New_York
-ENV C_FORCE_ROOT=true
+ENV TZ America/New_York
+ENV C_FORCE_ROOT true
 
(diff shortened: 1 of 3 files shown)


amcsz:
The specific PAM error message when a user is locked out is at line 65 at this link:
https://github.com/linux-pam/linux-pam/blob/master/libpam/pam_strerror.c

PR #1770, opened 2025-04-26
feat: Add code-based attendance option

### LINKED ISSUE #1612: Alternate attendance options
Ion should support more methods for taking attendance. Two main Ion-supported ways are currently used:
- Teacher calls and marks students 
- Students check off on printed attendance sheet

Other non-official ways include:
- Signing in by filling out a Google Form (especially for large activities)

---

Ideas for new attendance methods:
- Scanning a QR code/link -> takes students to a page that automatically marks them present
  - Keep track of time response is recorded (should be within block)
- Students input a code (that the teacher displays) into an attendance box on Ion
- ...more ideas welcome

TigerCodeHub:
Closes #1612
## Proposed changes
- Added a new method of attendance. For each EighthScheduledActivity, a random 6-character code 0-9 & A-Z is generated. This is displayed to the teacher (and eighth admins). Teachers can share the code to students at the activity, who can then input the code on a new eighth/student_attendance page which marks them as present if the code is correct (and shows a message as such).
- This also sets the attendance_taken on a sch. activity to True (as it would be odd otherwise).
- Note: Migration is required (new attendance_code attribute for EightScheduledActivity object).

- Attendance will only be marked if the code is given during the block (2:25 - 3:05 for A blocks, 3:15 - 4:00 for B blocks, anytime for other kinds of blocks). Otherwise, the student will get an error message.

- Students may access the attendance page through buttons which will appear on the eighth/glance (location.html) and eight/signup (signup.html) pages from 2:25 to 4:00.
- This attendance method coexists with already-existing methods (so the last change is the one which has effect).
- (All above times in PM).

This method (credits to Alan Zhu for the idea) will let students mark themselves present if they are at an activity (and thus receive the code from the teacher). This makes it more convenient for the teacher to take attendance, as all they need to do is show the code to the students at the activity.


alanzhu0 (changes requested):
Hi Tiger, great work so far. This is an excellent new attendance addition. Please see my comments, and also use pre-commit and get CI to pass.

I think a useful thing to add would be a checkbox to enable or disable the code. The teacher could enable the code and allow students to check-in, but then after everyone has done that, the teacher can disable the code and anyone who hasn't used the code yet can't do so after that. This is so that the teacher can keep an eye on who is using the code and when. This could replace the 2:25-3:05 restriction, actually. 

Also, in my issue I also mentioned an option for scanning a QR code or going to a link for taking attendance. If you could add that too, that would be great. I think that makes a lot of sense for going together with entering a code, and I think that they will have very similar logic flows. But that's up to you, and I won't make it a requirement - the work you've done here is more than enough.

TigerCodeHub:
Thanks for your feedback Alan!

I'll make a feature where the teacher can choose whether to manually enable/disable the code or let it run automatically based on times from the calendar (if they don't want to have to toggle it themselves). Also, I'll add the times from the calendar to the EighthBlock objects to record their start/end and make them more versatile (right now EighthBlock only stores date and letter - like A or B - which is why initially I hardcoded the code enable/disable times).

Your teacher-end pop up feature is a great idea. I'll try to put both the QR and character code in the pop-up together.

Thanks again,
Tiger

intranet/templates/eighth/location.html
```diff
@@ -43,6 +43,9 @@ <h5><strong>Sponsor(s):</strong> {{ s.3 }}</h5>
         </div>
         <br><br>
     {% endfor %}
+    {% if real_user.is_student and attopen%}
+        <a class="button" href="{% url 'student_attendance' %}">Take Attendance</a>
```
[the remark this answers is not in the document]
TigerCodeHub: Done!
(Also changed for signup.html).

intranet/apps/eighth/models.py
```diff
@@ -1195,6 +1202,15 @@ def set_sticky_students(self, users: "Sequence[AbstractBaseUser]") -> None:
                 bcc=True,
             )
 
+    def get_attendance_code(self):
```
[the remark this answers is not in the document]
TigerCodeHub: Ok I changed it to that.

Also, when migrating from an existing database, there's a one-time utility to run, because when migrating, the function is run once and that code is assigned to all the existing entries. To fix that, (as mentioned in the thread), I ran:

from intranet.apps.eighth.models import EighthScheduledActivity as esa
entries = esa.objects.all()
for e in entries:
    e.attendance_code = esa.random_code()
    e.save()

in the python shell.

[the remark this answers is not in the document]
JasonGrace2282: > Also, when migrating from an existing database, there's a one-time utility to run

You should write a custom migration if this is the case. Relevant docs:
https://docs.djangoproject.com/en/5.2/ref/migration-operations/#django.db.migrations.operations.RunPython
https://docs.djangoproject.com/en/5.2/topics/migrations/#data-migrations

intranet/apps/eighth/views/signup.py
```diff
@@ -215,13 +218,20 @@ def eighth_signup_view(request, block_id=None):
         except KeyError:
             active_block_current_signup = None
 
+        attendance_open = False
+        now = timezone.now().astimezone(pytz.timezone('America/New_York')).time()
+        a_start = tm(hour=14, minute = 25)
```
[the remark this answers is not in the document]
TigerCodeHub: Resolved in the second commit (I can squash them all together later).
I added this feature:
Sponsors (and admins) can toggle each EighthScheduledActivity between 3 modes: Auto, Open, and Closed. Auto will enable/disable the attendance code based on times drawn from the calendar model. Open will open it, and closed will disable it. Open/closed will override the times from the calendar.

TigerCodeHub:
Hi Alan,

I implemented the following features in the 2nd and 3rd commits:

- Sponsors/admins can choose between Auto/Open/Closed on their attendance page. Auto means the code is enabled 20 min prior to the block and closes 20 after the end of the block. Open/Closed will override the automatic times. 
- On the sponsor/admin attendance page, there is a pop-up button which displays a QR code and the attendance code. The QR code directs to a link which marks the student present, then redirects to the student attendance page with a success message. There are also more detailed instructions for the teachers.
- If a student's attendance has been marked, an "Attendance Marked" message will be shown on the student attendance page (so they know if they've done it already).

Some other things:
- Moved student attendance mechanisms to attendance.py from signup.py.
- The QR code URL contains both the scheduled_activity id and the attendance code, to ensure uniqueness
- Only sponsors/admin can see the code-based attendance things for an activity (as they are the only people who can take attendance the normal way).

Can you please review the new commits? (I've left them separate to keep it clear what's new, I can squash them together if needed.)

Thanks!
Tiger

alanzhu0 (changes requested):
Excellent work Tiger. This is looking great! I left a few small comments. I have not yet, but I or someone else will soon actually run and test out the code.

TigerCodeHub:
Hi Alan,
Thanks for your comments!

The QR code is now frontend-generated. Also, I made some minor changes (pulling the frontend rendering part of student_attendance_view into a new function to reduce function complexity, pulling the random_code function out of EighthSch..Activtiy to fix some errors in Codacy, etc.)

Tiger

intranet/apps/eighth/migrations/0072_auto_20250429_1128.py
```diff
@@ -0,0 +1,47 @@
...
+from django.db import migrations, models
+import intranet.apps.eighth.models
+
+def generate_attendance_codes(apps, schema_editor):
+    EighthScheduledActivity = apps.get_model('eighth', 'EighthScheduledActivity')
+    HistoricalEighthScheduledActivity = apps.get_model('eighth', 'HistoricalEighthScheduledActivity')
+    
+    for activity in EighthScheduledActivity.objects.all():
+        activity.attendance_code = intranet.apps.eighth.models.EighthScheduledActivity.random_code()
+        activity.save(update_fields=['attendance_code'])
```
aarushtools, line 12: This errors - random_code is not a function defined within EighthScheduledActivity?

intranet/apps/eighth/migrations/0072_auto_20250429_1128.py
```diff
@@ -0,0 +1,47 @@
...
+    
+    for historical_activity in HistoricalEighthScheduledActivity.objects.all():
+        historical_activity.attendance_code = intranet.apps.eighth.models.EighthScheduledActivity.random_code()
+        historical_activity.save(update_fields=['attendance_code'])
```
aarushtools, line 16: see above

intranet/apps/eighth/views/attendance.py
```diff
@@ -304,12 +307,20 @@ def take_attendance_view(request, scheduled_activity_id):
                 status=403,
             )
 
+        if "att_code_mode" in request.POST:
+            selected_mode = int(request.POST["att_code_mode"])
+            if scheduled_activity.code_mode != selected_mode:
+                scheduled_activity.set_code_mode(selected_mode)
+                redirect_url = reverse(url_name, args=[scheduled_activity.id])
+                return redirect(redirect_url)
+
```
aarushtools, line 316: couldn't this be simplified - I feel like there are unnecessary variables in your code here
```
attendance_code_mode = request.POST.get("att_code_mode")
if attendance_code_mode is not None and scheduled_activity.code_mode != int(attendance_code_mode ).. etc
```

intranet/apps/eighth/views/attendance.py
```diff
@@ -416,6 +430,7 @@ def take_attendance_view(request, scheduled_activity_id):
             "show_checkboxes": (scheduled_activity.block.locked or request.user.is_eighth_admin),
             "show_icons": (scheduled_activity.block.locked and scheduled_activity.block.attendance_locked() and not request.user.is_eighth_admin),
             "bbcu_script": settings.BBCU_SCRIPT,
+            "qrurl": request.build_absolute_uri(reverse("qr_attendance", args=[scheduled_activity.id, scheduled_activity.attendance_code])),
```
aarushtools, line 433: can you make it qr_url

intranet/apps/eighth/views/attendance.py
```diff
@@ -755,3 +770,126 @@ def email_students_view(request, scheduled_activity_id):
...
+                    invalidate_obj(act)
+                    attc = blk
+                except Exception:
+                    attf = blk
+                break
+            else:
+                attf = blk
+                break
+    return student_frontend(request, attc, attf, attimef, atteachf)
+
```
aarushtools, line 819: Reading this code is kind of difficult - can you make the variables longer in name / more clear so that others can understand it more easily on first read
Also, things like dayblks could be day_blocks since they're technically 2 words

aarushtools: oh and also type annotate your functions

intranet/apps/eighth/views/signup.py
```diff
@@ -215,13 +223,38 @@ def eighth_signup_view(request, block_id=None):
...
+        except Exception:
+            attendance_open = False
         context = {
             "user": user,
             "real_user": request.user,
             "block_info": block_info,
             "activities_list": safe_json(block_info["activities"]),
             "active_block": block,
             "active_block_current_signup": active_block_current_signup,
+            "attopen": attendance_open,
```
aarushtools, line 257: `att_open` instead of attopen. Or just expand it -> `"attendance_open": attendance_open`

intranet/templates/eighth/take_attendance.html
```diff
@@ -445,6 +440,81 @@ <h3>Passes</h3>
...
+                                            <input type="radio" name="att_code_mode" value="{{ value }}"
+                                                   {% if scheduled_activity.code_mode == value %}checked{% endif %}>
+                                            {{ label }}
+                                        </label>
+                                    {% endfor %}
+                                </form>
+                                &nbsp;&nbsp;<button type="submit">Save</button>
+
+                                <div class="activity-card">
+                                    <button type="button" onclick="showQRModal('{{ qrurl }}')">
```
aarushtools, line 469: Can you attach this event listener in the JS? It's generally bad practice to use inline event handlers (`onclick=`) for a variety of reasons

intranet/templates/eighth/take_attendance.html
```diff
@@ -445,6 +440,81 @@ <h3>Passes</h3>
...
+                                        Show QR Code
+                                    </button>
+                                </div>
+
+                                <div id="qrModal" style="display:none; position:fixed; top:0; left:0; width:100vw; height:100vh;
```
aarushtools, line 474: Can you use kebab case here and in other places where id="..."

intranet/templates/eighth/take_attendance.html
```diff
@@ -445,6 +440,81 @@ <h3>Passes</h3>
...
+                                        <div style="font-family: 'Fira Code', 'Consolas', monospace; font-size: 190px;">
+                                            {{ scheduled_activity.attendance_code }}
+                                        </div>
+
+                                        <div id="qrImageContainer" style="max-width:55000px; max-height:550px;"></div>
+                                        </div>
+
+                                        <br>
+                                        <button type="button" onclick="closeQRModal()">Close</button>
+                                    </div>
```
aarushtools, line 489: Same as above; attach the event handler in JS instead

aarushtools (changes requested):

TigerCodeHub:
Hi Aarush,

I did a bunch of fixes, mainly re-writing the process of passing the attendance-marking result, and adding a new attendance_marked boolean field to eighthsignup to record whether the student has done attendance so the student can see that (I couldn't use the was_absent field since it defaults to false and I didn't want to change that in case it affects something else), as well as some other things (like variable names and the javascript event listener).

Thanks for your feedback!
Tiger

aarushtools (changes requested):
This looks great! Just a few things:

1. Can students only enter attendance codes from the eighth/glance page (`eighth_location` view)? The glance page isn't easily accessible. How about add a little notification to the dashboard saying "The block you are in supports entering an attendance code"  or something along those lines only if the teacher has enabled attendance w/ code.

2. I know Alan linked to another qr code library, but can you use one that allows you to put your own logo in the middle of the QR. [Here's one](https://github.com/kozakdenys/qr-code-styling), but you can use another one you find if you'd like. Then put the Ion logo in the center of the QR code. You don't have to do this but it would look more stylish
3. When a teacher/admin clicks Show QR Code, can you add some instructions as to how the code should be entered in the popup itself for students to see
4. For the enter attendance code UI, can you move it out of glance and instead to its own page, and make it look somewhat like this
https://user-images.githubusercontent.com/1719915/159784302-4bb83708-e993-4800-9bcf-091ecb709ef7.png
(This is for 2FA but you get the idea)
5. Since you're moving the attendance code input part into its own page instead of being part of glance, the `qr_attendance_view` function can be removed entirely. Just make the new attendance code view take an optional GET parameter such as `?code=AAAA00`, and make the QR code link to the view's url with the GET parameter included, i.e. `/eighth/student-attendance?code=AAAA00`. Have JS automatically extract this code and then input it into the enter attendance code box and then auto-submit (all handled client-side).
6. Can you add docstrings to all your functions (see [here](https://github.com/tjcsl/ion/blob/23eff7fafacaf542613b0028c7d36ca05cedb6d3/intranet/apps/eighth/models.py#L49) and [here](https://github.com/tjcsl/ion/blob/23eff7fafacaf542613b0028c7d36ca05cedb6d3/intranet/apps/eighth/models.py#L1091) for good examples)

TigerCodeHub:
Hi Aarush,

To clarify, the attendance code UI is not part of the glance view, but rather another page which looks like glance (because I thought the format fitted well). I added buttons in the glance and eighth signup pages which directed to this page so students can access it (I can probably add some kind of notification/button into the dashboard as well). 

Do you think it would be better to have several pages (like one for each activity/block the student is signed up for in a day) for attendance input? Currently, its all in one page (the student_attendance_view / student_frontend / student_submit_attendance.html). I'll try to make the qr code process completely client-side.

Thanks for your detailed feedback!
Tiger

intranet/apps/eighth/models.py
```diff
@@ -1839,6 +1842,7 @@ def accept_pass(self):
         """Accepts an eighth period pass for the EighthSignup object."""
         self.was_absent = False
         self.pass_accepted = True
+        self.attendance_marked = True
         self.save(update_fields=["was_absent", "pass_accepted"])
```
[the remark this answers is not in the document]
TigerCodeHub: I think it would make sense to leave attendance_marked there (because since the pass is accepted, the student has basically been marked "present" and that should be reflected on the student end).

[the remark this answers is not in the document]
TigerCodeHub: Oh wait, I get what you mean.

TigerCodeHub:
Hi Aarush,
I implemented the following changes:

- Reworked the method of QR code attendance. Now, the QR code directs to the student_attendance_view with the two optional parameters filled, and the view (with helper functions) executes the attendance process and redirects to the generic student_attendance page (which clears the url of the params to prevent bugs). I think that if we already passed the code over in the url we could just execute attendance there rather than first passing it to client side then passing it back.
- Reworked the way how the time when global attendance opens/closes, by adding an open and close time fields to the DayType model, which is calculated by a method executed during save() or the migration. This means we don't have to calculate it every time a user access the relevant pages, which reduces load. 
- Added a setting which is the attendance code buffer time (currently 20 minutes).
- Added student instructions to the teacher end pop-up, and made the QR code more stylish (with the Ion logo and some coloring). Also displayed the "Auto" attendance open times to the teacher.
- The "View more blocks" button in the student eighth widget on the dashboard turns into "Enter Attendance Code" when attendance is open.
- Changed the code input box into 6 separate boxes.

Tiger

intranet/templates/eighth/take_attendance.html
```diff
@@ -445,6 +440,128 @@ <h3>Passes</h3>
...
+                                <b>Select Mode:</b>
+                                <form id="mode-form" method="post">
+                                    {% csrf_token %}
+                                    {% for value, label in scheduled_activity.mode_choices %}
+                                        <label>
+                                            <input type="radio" name="att_code_mode" value="{{ value }}"
+                                                   {% if scheduled_activity.code_mode == value %}checked{% endif %}>
+                                            {{ label }}
+                                            {% if value == 0 %}
+                                            {{ auto_time_string }}
```
aarushtools, line 464: Can you show this in 12 hr format

intranet/templates/eighth/location.html
```diff
@@ -43,6 +43,10 @@ <h5><strong>Sponsor(s):</strong> {{ s.3 }}</h5>
         </div>
         <br><br>
     {% endfor %}
+    {% if real_user.is_student and attendance_open%}
```
aarushtools, line 46:
```suggestion
    {% if real_user.is_student and attendance_open %}
```
sorry lol

aarushtools (changes requested):
Thanks Tiger! The UI is starting to look really good now. I like what you did with the QR code. Good thinking on reworking the DayType model; I think what you're trying to do right now would work perfect on a block-by-block basis instead. Keep in mind that by calculating the eighth window for each block individually instead of the day as a whole, the enter code box should only appear for the block that is open to the user right now.

Speaking of which, the QR code feature works great! The popup looks good now too. However, could you clarify instructions - where is the "attendance page"?? I'm still having trouble finding the page using just the UI - maybe I'm just being dumb, but I don't get a button anywhere to "Enter Attendance Code" except for glance? Could you send screenshots of where they are. I don't think it's rendering since I Ctrl + F the whole HTML and still couldn't find it.

For entering the actual attendance code, can you move it out of the glance-looking page and into its own page (and each block has a separate page) where everything is bigger. The 6 digit code thing works great when typing + backspacing, but can you make the boxes perfect squares? If it looks weird you can use a different font in the boxes themselves to make it look better. Also, even though attendance through a specific URL with the code works (such as the one provided by the QR code), trying to enter it manually and hitting submit just does reloads the page and does nothing. I noticed this only happens for A block (works for B block). I also noticed that when I type in the 6 boxes for B block, the cursor doesn't automatically move to the next box - but you probably won't need to worry about this as I want you to move each block to its own attendance code page.

Other than that, this is looking really good! Sorry for all these nitpicks about how it looks and such, but since this will probably be a widely used feature I want to make sure it looks and works well for the user

aarushtools:
Also can you make this say `Enter Code →` or something along those lines if the activity they've signed up for has code attendance open
image

TigerCodeHub:
Hi Aarush,
Thank you for your really detailed feedback!

Some clarifications/questions I have:
The "Enter Attendance Code" not appearing is odd (it appeared on my end in the eighth signup view - like if you click on a block to see the activity you signed up for - and on the upper right of the dashboard). Both that and the attendance itself are drawn from the schedule using the same logic, so its weird that attendance works but that doesn't. I'll test it some more, and try to get some screenshots.
There might have been a bug when I moved some of the code attendance logic around, so I'll try to fix that. Perhaps it errored for A block but didn't show the error message? 
Also, the problem with calling super() after generating is that the generation depends upon the DayType's blocks, and if they were modified/created it's best to save them first then reference them (unless it's fine to not do that?) Instead of save I'll probably use update instead. Also I can definitely move this logic to Block. 

Tiger

aarushtools (changes requested):
Hi Tiger,

Sorry for the late review - and this is really only a partial review for now - but a few thoughts:
- There are some padding issues (you may already know about this)

Image
(Another example: the J)

Image
(otherwise the 2fa box looks and works great now, good job!)

- Can we unify the enter_student_attendance page to NOT be dependent on the block? This may take some logical rewriting, but I believe it should make the logic simpler overall. What I mean by this is that there is one page for entering codes indifferent to parametric route arguments, and that upon entering a code, you reverse lookup that code to find the block and activity it is for. This means that all codes should be guaranteed unique. You can probably reduce the slim chance of collisions by scheduling a celery task to delete old codes every so often. 

- The QR code idea is good, but I think it makes more sense to pass the code in as a query param: `http://localhost:8080/eighth/student_attendance?code=J8J7YK` instead of `http://localhost:8080/eighth/student_attendance/1/J8J7YK` (ignoring the /1 because if the above is implemented block id shouldn't matter)

- This is up to you and we can discuss it later, but it seems like a lot of logic is wrapped up in the "Auto" mode. I think we can get rid of it. Teachers should be expected to manually enable/disable anyways - they don't even know when "auto" starts/stops.

- I know this was my idea, but if the enter_student_attendance page is unified to be independent of the block id, then it might make sense to remove the "Enter Code" thing for each block

Image

Instead, somewhere in the red, you can put a big banner saying "The activity `___` you are in has code attendance enabled. Enter it here ->"

- Don't show the attendance code if code attendance is closed. It should start at closed by default.

- As students mark their attendance, the teacher has to refresh the page to see it happen. This is once again up to you, but do you think a websocket approach to update it live is better (and is it worth the extra effort)?

Great job so far and keep up the good work!

TigerCodeHub:
Superseded by #1926. The attendance work has been moved from my dev branch to a dedicated ttendance branch so my fork's dev can stay in sync with upstream. Continuing there — thanks for the reviews here!

PR #1771, opened 2025-04-26
feat(printing): added descriptive user input errors for page_range

### LINKED ISSUE #1717: More descriptive errors for printing page rage
## Topic of discussion
The page range option on the Ion printing page does not clearly describe the possible ways to represent the range.
For example, it might be intuitive for some people that to get the first 3 pages the page range might be `0-3`, but pages
aren't zero indexed. Furthuremore, Ion only gives the vague error message of an invalid page range, which doesn't tell the user if it was a parsing issue, or an out of bounds issue, or something else.

## Proposed options/solutions
Rewrite `check_page_range` https://github.com/tjcsl/ion/blob/8cb77914c830da3b47ff812f46421b526b9fbf76/intranet/apps/printing/views.py#L244
to return more descriptive errors (instead of returning `None`), and maybe add a `help_text` for the page range option.

Whatever behavior is added should also be tested thoroughly.

## Personal opinion (if applicable)
This is probably a good first issue for contributors.

It also might be worth considering making the `check_page_range` function a form field validator and/or overriding `Form.clean` - see https://docs.djangoproject.com/en/4.2/ref/forms/validation/

sroshc:
Closes #1717 

## Proposed changes
- Add a validator to the page_range field in PrintJobForm to serve descriptive syntactical error messages to the user
- Add a help text to the page_range field 
- Refactor check_page_range and print_job to correctly inform the user if their pages are out of bounds

Users can clearly see how to format a page range and get descriptive errors about the syntax. Also, users are now informed if their page range exceeds the amount of pages in the inputted document.


intranet/apps/printing/forms.py
```diff
@@ -25,13 +25,64 @@ def validate_size(self):
...
+        """
+        Expects string that only includes commas, digits, and dashes (already validated by regex)
+        Validates that the page_range field's ranges and numbers correctly selects any pages from 1 to infinity.
+        """
+        # Return if user specifies all pages.
+        if self.strip() == "":
+            return
+
+        # Gets number with suffix for clean error handling, e.g 1st, 2nd, 3rd
+        def get_number_ordinal(n: int) -> str:
```
aarushtools, line 38: nice

intranet/apps/printing/forms.py
```diff
@@ -25,13 +25,64 @@ def validate_size(self):
...
+                if len(input_range) != 2:
+                    raise forms.ValidationError(error_prefix + "make sure the range is properly formatted (x-y).")
+
+                try:
+                    range_from, range_to = map(int, input_range)
+                except ValueError as e:
+                    raise forms.ValidationError(error_prefix + "you must specify two numbers around a dash for ranges.") from e
+
+                if range_from <= 0 or range_to <= 0:
+                    raise forms.ValidationError(error_prefix + "there is no 0th page.")
```
aarushtools, line 62: Make it a little more clear, maybe something along the lines of "Pages are numbered starting at 1, not 0"

intranet/apps/printing/forms.py
```diff
@@ -25,13 +25,64 @@ def validate_size(self):
...
+
+                if range_from > range_to:
+                    raise forms.ValidationError(error_prefix + "the lower bound should be larger than the upper bound.")
```
aarushtools, line 65: Did you mean to say that the lower bound should be smaller than the upper bound?

aarushtools (changes requested):
This looks great overall, just some small changes need to be made regarding the messages. Also, we might need to test this on [snowy/iondev](iondev.sites.tjhsst.edu) since testing printing features is hard on the dev env, unless you already manually tested it yourself? We can do this during understudy 8th. cc @alanzhu0

intranet/apps/printing/forms.py
sroshc: I did, i'll fix that asap

sroshc:
I have tested the individual validator and check_page_range function and they both should behave as expected. I also checked if `lpr` accepts the range outputs the validator sees as correct on my own machine using a dummy printer. I'll make the changes to the messages soon as well.

sroshc:
Squashed the original changes and the error message fixes together btw

aarushtools (approved):

sroshc:
I'll test it out during lunch on noonday, for this I would need to be connected to the school's network right?

sroshc:
I tested a few prints, and the error handling works as intended, but some valid prints get denied by `lpr`. `lpr` only accepts print ranges in ascending order. For example, "1-2, 5" will be accepted but "5, 1-2" will not be accepted by `lpr` and it will error out with "lpr bad range". This error is present on production and is not handled. I will make changes to my pr addressing this and test it after.

sroshc:
I have added changes addressing the ion bug of unhandled `lpr` exceptions. Also, CUPS ignores if a page is specified multiple times, and now the user is informed instead of it originally ignoring this. I have tested the validator, and this should work fine as that is the only part of the code I edited. If you feel like this requires testing, I can test it during any lunch period.

shrysjain (approved):
Nice work, LGTM

@alanzhu0 ready to merge

PR #1780, opened 2025-05-06
fix: signage bus format

TigerCodeHub:
## Proposed changes
- Fixes the format on the bus displays in signage. It moves the bus grid to the left (by changing the bus container div width to 50% of page - and html automatically left-aligns), avoiding overlap with the announcements.

This prevents the announcements and "School is in session/..." sign from obscuring the buses. 

Top is modified signage, bottom is current signage.
image
image

### DIFF
--- a/intranet/templates/signage/pages/bus.html
+++ b/intranet/templates/signage/pages/bus.html
@@ -357,7 +357,7 @@
         </script>
         {% endverbatim %}
         <div class="primary-content">
-            <div class="info bus-signage">  </div>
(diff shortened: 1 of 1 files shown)


aarushtools (changes requested):
Thanks Tiger - one small thing, the gap is a little too big. Can you make the width 65% instead?

aarushtools (approved):

TigerCodeHub:
I think the TV is also affected. Below is Hopper commons signage during dismissal today:
IMG_0366_compressed

PR #1783, opened 2025-05-09
fix(announcements): display events correctly on announcement archive

### LINKED ISSUE #1775: Empty pages in announcement archive
Currently, the [announcement archive](https://ion.tjhsst.edu/announcements/archive) has a lot more pages than needed for archived announcements leaving many empty pages with no announcements on them. Figure out why this is and fix the pagination to work as intended.

The issue might have potentially started with [4b10afc](https://github.com/tjcsl/ion/commit/4b10afc6584aead7b0553295d9a2b38b8072e1f0) so I'd recommend starting with looking at that commit.

I doubt its going to be a huge thing and most of the work will probably be debugging. Good luck :)

makors:
Closes #1775. Issue surfaced in eb3cbd9 and hasn't been seen previously as there weren't enough events for the issue to be noticeable.

## Proposed changes
- Filter `Event` objects on dashboard view to `show_on_dashboard=True` for requests with `show_expired` set to True

The announcements archive page is not supposed to show events through its filters. Before, blank HTML items were in place of the events, which are no longer rendered with this filter.

### DIFF
--- a/intranet/apps/dashboard/views.py
+++ b/intranet/apps/dashboard/views.py
@@ -256,7 +256,7 @@ def get_announcements_list(request, context) -> list[Announcement | Event]:
     if context["events_admin"] and context["show_all"]:
         events = Event.objects.all()
     elif context["show_expired"]:
-        events = Event.objects.visible_to_user(user)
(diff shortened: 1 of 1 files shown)


aarushtools (approved):
Good work!

PR #1785, opened 2025-05-15
feat: delete old lostfound entries

TigerCodeHub:
## Proposed changes
- Created a task which deletes lost and found entries more than two months old. Scheduled celery beat to run it every month.
- Changed the display order of lost and found entries to most recent on top (as it is usually easier to find recently lost items). (Note: I switched to .all() because the item model meta class already has ordering = ["-added"]).
- Added a link to the lost and found page in the Useful Links -> TJ Resources section (and removed the Webmail and Schoology links in the process, since Webmail is usually accessed from the webapps dropdown and Schoology is so common that this link isn't really necessary).

I noticed there was an inactive lost and found feature in Ion with some really old entries. I think it could serve as a catalog (like we could encourage people who found items and put them in the physical lost&found to also enter it on Ion so the owner could know that the item was found). Clearing old entries helps to reduce clutter and encourage people to use the page (since they can then get new information).


intranet/templates/dashboard/links.html
```diff
@@ -29,8 +29,7 @@ <h5 class="link-heading">Tutorials</h5>
         </td>
         <td style="width: 50%">
           <h5 class="link-heading">TJ Resources</h5>
-          <a href="https://webmail.tjhsst.edu/">Webmail</a><br>
-          <a href="https://lms.fcps.edu/">Schoology</a><br>
+          <a href="https://ion.tjhsst.edu/lostfound">Lost and Found</a><br>
           <a href="https://studyguides.sites.tjhsst.edu">SGA Study Guides</a><br>
           <a href="https://tutoring.sites.tjhsst.edu/">TJ Peer Tutoring</a><br>
```
aarushtools, line 34: While you're at it, I think you can also get rid of the tutoring url since it doesn't exist anymore, and then move the changes for this file into a separate commit

intranet/apps/lostfound/tasks.py
```diff
@@ -0,0 +1,17 @@
...
+from django.utils import timezone
+
+from .models import FoundItem, LostItem
+
+logger = get_task_logger(__name__)
+
+
+@shared_task
+def remove_old_lostfound():
+    expired = timezone.now() - timedelta(days=60)
```
aarushtools, line 14: can you add the number of days before clearing as a setting [here](https://github.com/tjcsl/ion/blob/23eff7fafacaf542613b0028c7d36ca05cedb6d3/intranet/settings/__init__.py)

aarushtools: Also, can you bump up the cooldown to 180 days (as the default value for the setting). The lost and found feature barely gets used so it won't matter much anyway

aarushtools (changes requested):
Thanks for the PR Tiger. One small thing though - can you hide the items instead of deleting them? Just in case they need to be recovered manually in the future

TigerCodeHub:
Hi Aarush
Thanks for your feedback! I changed the code accordingly (I added a new expired field since it wouldn't make sense to call an item "found" or "retrieved" if it were just expired.)
Tiger

TigerCodeHub:
Wait hold on, git is being a little funny in moving code around commits. Let me fix that.

aarushtools (approved):
Thanks!

PR #1791, opened 2025-05-17
fix: restyle status highlight and normalize container height

YousafZahid1:
Let me update the description with the correct files you modified:

Closes #1789. Issue wasn't previously noticeable until it was identified in the navigation bar styling.

## Proposed changes
- Fixed status text highlight in collapsible navigation bar to be round and darker, similar to header status text
- Adjusted container box size for status text on bus page to match other pages, correcting the nav height issue

The status text in the collapsible navigation bar had a full white highlight on click instead of the round, darker highlight used elsewhere. Additionally, the container box for status text was oversized on the bus page due to a generic selector in the bus scss, causing inconsistent nav height across pages.

Changes made in:
- intranet/templates/page_with_header.html
- intranet/static/css/bus.scss

Before:
Screenshot 2025-05-17 at 2 31 13 PM

After:
Screenshot 2025-05-17 at 2 30 06 PM


intranet/static/css/bus.scss
```diff
@@ -203,7 +203,7 @@ svg:not(.status-icon-svg) {
 }
 
 h2.status {
-    margin-bottom: 5px;
+    margin-bottom: 0px;
```
aarushtools, line 206: I don't see how this fixes the issue

aarushtools (changes requested):
Thanks for the PR Yousaf, but it seems like your changes cause the existing highlight behavior to malfunction. Look at the screenshots below:

Your branch:
image
image

Should be:
image
image

Also, clicking on the status text gives a good highlight, but it shouldn't shift the entire div by a few pixels.

PR #1793, opened 2025-05-18
fix: restyle status highlight and normalize container height

YousafZahid1:
Let me update the description with the correct files you modified:

Closes #1789. Issue wasn't previously noticeable until it was identified in the navigation bar styling.

## Proposed changes
- Fixed status text highlight in collapsible navigation bar to be round and darker, similar to header status text
- Adjusted container box size for status text on bus page to match other pages, correcting the nav height issue

The status text in the collapsible navigation bar had a full white highlight on click instead of the round, darker highlight used elsewhere. Additionally, the container box for status text was oversized on the bus page due to a generic selector in the bus scss, causing inconsistent nav height across pages.

Changes made in:
- intranet/templates/page_with_header.html
- intranet/static/css/bus.scss

Before:
Screenshot 2025-05-17 at 2 31 13 PM

After:
Screenshot 2025-05-17 at 2 30 06 PM


intranet/static/css/page_base.scss
```diff
@@ -511,7 +511,7 @@ ul.dropdown-menu.absence-notification {
     }
 
     .selected {
-        background-color: rgb(252, 252, 252);
+        background-color: rgb(223 222 222);
```
aarushtools, line 514: You're still changing .selected so the same behavior of broken highlighting will occur, just in another color

aarushtools (changes requested):

intranet/static/css/page_base.scss
YousafZahid1: Just to confirm, the error you're talking about is when you click on the link and then go back to ion the effect isn't there, and that's what I need to resolve?

aarushtools: No, when you click and then go back to ion a weird white highlight appears. The problem with changing .selected __for all elements__ is that it affects everything clicked in the navbar (look at the pictures I sent above). Just test it out - you shouldn't change the existing behavior of the other navbar elements in anyway

YousafZahid1: Can you provide with a image this is a pr

intranet/static/css/page_base.scss
```diff
@@ -521,6 +521,9 @@ ul.dropdown-menu.absence-notification {
     .nav-status-indicator {
         margin-top: 10px;
```
aarushtools, line 522: move the .selected marker inside .nav-status-indicator so it won't affect other elements

intranet/static/css/page_base.scss
```diff
@@ -521,6 +521,9 @@ ul.dropdown-menu.absence-notification {
...
         display: none; // only show when the nav bar is collapsed
+        border-radius: 14px;
+        width: 187px;
+        height: 27px;
```
aarushtools, line 527: Make this have the same border-radius as the header status' highlight border radius

intranet/static/css/page_base.scss
aarushtools: see my comment below

intranet/static/css/page_base.scss
```diff
@@ -510,20 +510,22 @@ ul.dropdown-menu.absence-notification {
...
 
     .nav-status-indicator {
+
         margin-top: 10px;
         display: none; // only show when the nav bar is collapsed
-        border-radius: 14px;
+        border-radius: 8px;
         width: 187px;
         height: 27px;
+        &.selected {
```
aarushtools, line 522: This works, but keep the old .selected as well on line 513.

intranet/static/css/page_base.scss
```diff
@@ -510,20 +510,22 @@ ul.dropdown-menu.absence-notification {
...
+            background-color: rgb(223 222 222); // Apply the background color to the .nav-status-indicator itself when selected
+
+            a {
```
aarushtools, line 525: Remove this, the link should be blue when you click on it

aarushtools (changes requested):
The status text and icon in the background highlight is off-center. Can you fix this as well? Thanks

image

intranet/static/css/page_base.scss
```diff
@@ -510,20 +510,22 @@ ul.dropdown-menu.absence-notification {
...
-    .selected {
-        background-color: rgb(223 222 222);
 
-        a {
-            color: $grey;
-        }
-    }
 
     .nav-status-indicator {
+
```
aarushtools, line 516: get rid of this extra line

YousafZahid1:
Screenshot 2025-05-19 at 8 22 59 PM

intranet/static/css/page_base.scss
```diff
@@ -521,6 +520,25 @@ ul.dropdown-menu.absence-notification {
...
+        height: 27px;
+        margin-left: auto;
+        margin-right: auto;
+        &.selected {
+            background-color: rgb(223 222 222);
+            a {
+                color: $grey;
+            }
+        }
+        &.selected a {
```
aarushtools, line 536: You can add all these properties on line 532. Sass compiles `&.selected { a {} }` as `&.selected a {}`

aarushtools:
Also squash your commits - you don't have to make a new PR for this, use `git push --force` to push your changes if you get a rejected error.

intranet/static/css/page_base.scss
```diff
@@ -509,19 +509,31 @@ ul.dropdown-menu.absence-notification {
...
         }
     }
 
     .nav-status-indicator {
         margin-top: 10px;
-        display: none; // only show when the nav bar is collapsed
+        display: flex;
+        align-items: center;
+        justify-content: center;
+        border-radius: 8px;
```
aarushtools, line 524: It's still not justified center, it's off slightly. Look closely and use inspect - lines 521-523 can be removed to actually align it in the center correctly.
Also, increase the background highlight height - Give 3-4px clearance with the icon.

aarushtools (changes requested):

aarushtools (changes requested):
It's still too big on the bus page compared to other pages (see the original issue)
image

Also, squash commits. Once you fix the above issue this should be good to go.

aarushtools (changes requested):
You solved the height issue, but now it's not centered vertically anymore (a little too high)
image

intranet/static/css/page_base.scss
```diff
@@ -520,8 +519,22 @@ ul.dropdown-menu.absence-notification {
...
-
+        display: none;
+        border-radius: 8px;
+        margin-left: auto;
+        margin-right: auto;
+        height: 23px;
+        padding: 3px 4px;
+        &.selected {
+            background-color: rgb(223 222 222);
+            vertical-align: center;
```
aarushtools, line 530: https://developer.mozilla.org/en-US/docs/Web/CSS/vertical-align

`center` is not an appropriate value for `vertical-align` 
image

Also, according to the inspect, background-color doesn't do anything

aarushtools (changes requested):

intranet/static/css/page_base.scss
aarushtools: in the screenshot I moved vertical-align outside of .selected but it's the same effect - it should be outside of .selected anyways right? Or did you mean to only apply vertical-align: middle when it's selected on purpose

YousafZahid1: Oops — I forgot to remove background-color: rgb(252,252,252);. Also vertical-align: middle; is correct, it didn’t have any effect, so I’ll take that out too.

aarushtools (approved):

aarushtools:
oops squash commits though

intranet/static/css/page_base.scss
```diff
@@ -520,8 +518,22 @@ ul.dropdown-menu.absence-notification {
 
     .nav-status-indicator {
         margin-top: 10px;
-        display: none; // only show when the nav bar is collapsed
```
aarushtools, line 523: add this comment back

aarushtools (changes requested):
``fix(nav)``

PR #1797, opened 2025-06-05
fix(docs): update usernames.md and update reference

makors:
## Proposed changes
- Update users in `docs/source/developing/usernames.md` to match the current year and add a footnote describing the calculations made for new users (usernames will have to be updated yearly though)
- Updated reference to `usernames.md` which was changed in 6f591f1

The documentation does not currently reflect the correct information for logging into a development environment, which is of course required to use Ion in a development environment.


aarushtools (approved):
Thanks!

PR #1820, opened 2025-09-05
fix(eighth): fix attendance form autosubmission

hstievat:
## Proposed changes
- Fix jquery selector and re-comment out next buttons

-


aarushtools (approved):
Thanks!
note: Resolves #1817 and #1819

PR #1830, opened 2025-09-14
fix(ui): consistency when rejecting/approving events

### LINKED ISSUE #1825: Fix UI consistency when rejecting/approving events
## Steps to reproduce
1. Create an event
2. Click on the events tab
3. Click show all
4. Reject/approve the announcement

## Expected behavior
When the event is rejected/approved all of the events are still shown, like when you initially clicked the show all button

## Actual behavior
When the event is rejected/approved, you have to click the show all events button again as they automatically collapse.

Fix: respect any query params (`?=...`) in the url

ZamanZahid:
closes #1825 
## Proposed changes
- keeps ?show=all, after approve/reject

### DIFF
--- a/intranet/static/js/announcement.form.js
+++ b/intranet/static/js/announcement.form.js
@@ -143,11 +143,11 @@ $(function() {
         exp.val(dateFormat(new Date($(this).data("date"))));
     });
 
-    $("#date-reset-btn").click(function () {
+    $("#date-reset-btn").on("click", function () {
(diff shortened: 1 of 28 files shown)


aarushtools (changes requested):
This solution is innovative, but can you use `request.GET.urlencode()` and just add that to the url instead. It's more verbose and http_referer could have bad input. Thanks

PR #1848, opened 2025-10-28
Fix postgres from crashing

amcsz:
## Proposed changes
See diff

New postgres version came out recently, which broke the docker compose. Changing `/var/lib/postgresql/data` to `/var/lib/postgresql` fixes this.


amcsz:
oops I included a past commit

PR #1850, opened 2025-10-29
feat(emerg): move csl status checker to celerybeat task

### LINKED ISSUE #1818: Make status checker run on a separate thread
## Feature description
Change the Ion CSL status checker (added in #1754 ) to run on a separate thread instead of the worker thread.

Currently, the CSL status checker will start a request to [https://status.tjhsst.edu/index.json](https://status.tjhsst.edu/index.json) if the 5-minute cache has expired. This process can take up to 30 seconds, especially if the initial request has timed out (which happens often). Additionally, if multiple people make requests to Ion at the same time, and the cache has expired, multiple requests will be sent to that same url, causing a page hang for many users simultaneously.

- Use a celery task to check the status every 5 minutes with celerybeat. The cache should NOT expire every 5 minutes so that requests to Ion while the status is being updated still go through.

amcsz:
## Proposed changes
- Title
- CSL Status cache expires in 10 minutes instead of 5
- Task will occur in background instead of on user demand
- Status is still stored in cache

Closes #1818


intranet/apps/emerg/views.py
```diff
@@ -123,38 +121,14 @@ def update_emerg_cache(*, custom_logger=None) -> None:
...
-            status = response.json()["data"]["attributes"]["aggregate_state"]
-            updated = True
-        except Exception as ex:
-            status = "error"
-            logger.error(f"Could not fetch status page or incorrect status page JSON format: {ex}")
-
-        cache.set("emerg:csl_status", status, settings.CACHE_AGE["csl_status"])
-
-    return status, updated
+    return cache.get("emerg:csl_status")
```
aarushtools:
```suggestion
    return cache.get("emerg:csl_status", "error")
```
to stay consistent with returning only one of 5 values, else we might accidentally return None

aarushtools (changes requested):
Thanks for this! Just 2 small changes
Also, can you add closes #1818 to your commit

aarushtools (changes requested):
Sorry one more thing, when the server is restarted celery should immediately run the task. Currently, it does so after 5 minutes, so the Ion homepage returns "unknown status" until then.

PR #1866, opened 2025-12-23
feat(eighth): Allow 8th period activities to be scheduled but not seen

ElijahFeldman7:
## Proposed changes
- title
- 
- 

closes #1863


shrysjain (changes requested):
Hi Elijah, good work so far! A couple changes to make:

Right now, when an activity is marked as hidden it does not show up for students in the eighth signup page, but the block can still be seen from the activity page (`eighth/activity/<id>`). Through this students can also sign up for hidden activities which defeats the purpose. Can you make it so the scheduled block is also not visible here until the appropriate date has passed?

image

*This block has the activity marked as hidden and is not visible from signup but is visible from the activity page when logged in as a student*

Also, the functionality for adding a "hidden until" date in eighth admin (`eighth/admin/scheduling/schedule`) seems to work, but after clicking the save button the date no longer shows in the text area whereas things like comments, checkmarks, e.t.c. do persist. Can you figure out why this is happening and fix that as well. 

Screenshot 2025-12-23 at 10 42 43 PM

*I set a hidden until date and saved but it does not show*

Other than that, great work and this should be pretty close to being complete :)

aarushtools (changes requested):
Hi Elijah,
Good work on this issue so far, just a few more things on top of what Shreyas said:
image
Can you format this tooltip text to separate the date and time better (and on a 12 HR clock). Also, the hidden badge has the same color as 2-day - can you make it have a unique styling (but use bland colors)
Also, can you prevent add_user from going through if the activity is hidden

ElijahFeldman7:
ok this should be good now; i made the icon grey btw

intranet/apps/eighth/models.py
```diff
@@ -385,6 +385,14 @@ def get_active_schedulings(self) -> QuerySet | Collection["EighthScheduledActivi
...
         return EighthScheduledActivity.objects.filter(activity=self, block__date__gte=date_start, block__date__lte=date_end)
 
+    @property
+    def is_hidden(self) -> bool:
+        """Returns whether the activity is hidden.
+        Returns:
+            Whether the activity is hidden.
+        """
+        return self.get_active_schedulings().filter(hidden_until__isnull=False, hidden_until__gt=timezone.now()).exists()
+
```
aarushtools: Why did you put this on normal activity shouldn't it just be on scheduled activity?

intranet/apps/eighth/models.py
```diff
@@ -1222,6 +1239,10 @@ def add_user(
 
         exception = eighth_exceptions.SignupException()
 
+        if self.is_hidden:
```
aarushtools: good

intranet/apps/eighth/models.py
aarushtools: I see you're using it in activities.py line 38 but is this usage necessary or can it just be replaced by a model query? If it is I would rename this property to make it more clear i.e. has_any_hidden_schedulings? and perhaps make it not a property to indicate this runs a somewhat complex query

aarushtools (changes requested):
Hi Eli, just a few quick changes. Also can you make a 2nd commit while you're at it since it's similar:
- disables seeing administrative AND hidden activities using `/eighth/signup?activity={activity_id}` url if you don't have 8th admin (basically if you can't see it normally in the activity list)

aarushtools:
Also I can see hidden activities on `/eighth/activity/{activity_id}` when I shouldn't be able to
edit: shreyas mentioned this, are you sure you fixed it?

PR #1869, opened 2025-12-25
feat(profile): counselor dropdown instead of entering in id

ElijahFeldman7:
## Proposed changes
- title
- 
- 

closes #1861


aarushtools:
Thanks Eli, we put this in #1929 since we added one more dropdown there

PR #1873, opened 2026-01-08
feat(auth): implement cloudflare turnstile on login page

makors:
## Proposed changes
- Adds a Cloudflare Turnstile widget to the login/auth form with proper validation checks
- CSS/JS changes for functionality and to allow for the Turnstile widget to fit if it's visible (typically invisible, sometimes visible. see [managed mode](https://developers.cloudflare.com/turnstile/concepts/widget/))
- Adds tests for the CAPTCHA solution and updates current ones accordingly with cases for enabled and disabled CAPTCHAs

This increases the cost of brute forcing Ion with logins (proxies are pretty easy) with CAPTCHAs in addition to the current login security measures.


makors:
side note: in MOST CASES the widget will be invisible under managed mode (assuming a moderate reputation), but the test keys used in this pr are only for the visible widget, so they'll have to be changed to the commented out keys for testing in settings/__init__.py (secret key is set to always pass).

makors:
big ci wants you to think you can't commit secrets

makors:
Closed in favor of #1888

PR #1876, opened 2026-01-14
fix(tests): fix local development tests through docker

makors:
## Proposed changes
- Update Ion documentation to reflect updated local testing through various methods
- Fix OAuth Settings to work locally with pytest
- Fix deprecated config setting

The people need to test!!!

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -47,7 +47,7 @@
                 },
                 {
                     "name": "Set up packages",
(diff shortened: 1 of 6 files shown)


aarushtools (changes requested):
Hi Bryce,

You actually bring up a good point about pytest docs being out of date. To my knowledge, Ion doesn't use pytest meaningfully anywhere, since we opt to use Django's test runner instead (even in CI). Can you actually just remove any references to pytest, including in dependencies, docs, etc. The other changes are good.

aarushtools (approved):
Thanks Bryce!

PR #1887, opened 2026-04-10
fixed (enrichment): default to next week on weekends

### LINKED ISSUE #1875: Default to next enrichment week on Saturday
On saturday and sunday, the past week´s enrichment activities are default render. These cannot be signed up for after friday, so it would be more useful to switch to the next week´s enrichment activities on saturday.

ZamanZahid:
closes: https://github.com/tjcsl/ion/issues/1875

## Proposed changes
*  Added helper function to determine default enrichment date
* Updated enrichment_context to use new weekend logic

Right now, on Saturdays and Sundays, the enrichment page defaults to the last week's activities, even though signups are closed after Friday. This change makes it show the next week instead, unless there are still enrichment activities on that weekend.

### DIFF
--- a/intranet/apps/enrichment/views.py
+++ b/intranet/apps/enrichment/views.py
@@ -36,11 +36,33 @@ def is_weekday(date):
     return date.isoweekday() in range(1, 6)
 
 
-def enrichment_context(request, date=None):
+def get_default_enrichment_date(request):
     local_time = timezone.localtime()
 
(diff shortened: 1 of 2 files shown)


aarushtools (changes requested):
You need to place this commit on a new branch

ZamanZahid:
Alright here is the new one -  https://github.com/tjcsl/ion/pull/1899

PR #1890, opened 2026-04-22
fix(polls): preserve question editor state on form validation error

### LINKED ISSUE #1843: Don't make polls questions reset when form error occurs

CrepuscularIRIS:
## What

Preserve the poll question editor's content when an add-poll or modify-poll form fails validation, so users don't have to re-enter their questions.

## Why

Fixes #1843

When a user fills in poll questions and submits the form but a validation error occurs (e.g. missing end time), the page re-renders with the question editor empty. In `add_poll_view` the context always passes `poll_questions: "[]"` and `poll_choices: "[]"` on every render, including error re-renders. In `modify_poll_view` the context uses the DB state, which doesn't reflect the user's unsaved edits. In both cases the dynamically built question editor loses all user input.

## How

Pass the submitted `question_data` string back to the template as a new context variable `submitted_question_data`. If non-empty (i.e. we're re-rendering after a POST error), `polls.js` reinitialises the question editor from that data instead of starting from scratch.

The change is in three small parts:
1. `views.py` — add `"submitted_question_data": request.POST.get("question_data", "[]")` to the render context in both `add_poll_view` and `modify_poll_view`
2. `add_modify.html` — expose the variable to JavaScript alongside the existing `poll_questions`/`poll_choices`
3. `polls.js` — after the normal initialisation from `poll_questions`/`poll_choices`, check whether `submitted_question_data` is non-empty and rebuild the editor from it

Alternatives considered: converting `question_data` back into Django-serialiser format (`poll_questions`/`poll_choices`) in Python. This works for the modify path (questions have PKs) but fails for the add path (no PKs yet), so a uniform JS approach was cleaner.

## Scope

- Included: add-poll form and modify-poll form, both POST-error paths
- NOT included: the vote form or any other form in the polls module

## Verification

- [x] Two regression tests added (`test_add_poll_preserves_questions_on_form_error`, `test_modify_poll_preserves_questions_on_form_error`) that assert `submitted_question_data` is present in the response context and contains the correct question data when the form fails validation
- [ ] Manual check: submit an add-poll form with questions but an invalid end time — questions should remain in the editor after the error message appears
- [x] Existing polls test suite should continue to pass

## AI Disclosure

AI tools were used to assist with implementation. Every line was reviewed for correctness and intent.

### DIFF
--- a/intranet/apps/polls/tests.py
+++ b/intranet/apps/polls/tests.py
@@ -478,9 +478,8 @@ class ApiTest(IonTestCase):
 
         self.assertEqual(response.status_code, 200)
         self.assertIn("submitted_question_data", response.context)
-        import json as _json
 
-        restored = _json.loads(response.context["submitted_question_data"])
+        restored = json.loads(response.context["submitted_question_data"])
         self.assertEqual(len(restored), 1)
         self.assertEqual(restored[0]["question"], "Favourite colour?")
         self.assertEqual(len(restored[0]["choices"]), 2)
@@ -516,8 +515,7 @@ class ApiTest(IonTestCase):
 
         self.assertEqual(response.status_code, 200)
         self.assertIn("submitted_question_data", response.context)
-        import json as _json
 
-        restored = _json.loads(response.context["submitted_question_data"])
+        restored = json.loads(response.context["submitted_question_data"])
         self.assertEqual(len(restored), 1)
         self.assertEqual(restored[0]["question"], "Updated question?")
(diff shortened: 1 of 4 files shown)


CrepuscularIRIS:
Just pushed a small follow-up: added a `/* global */` declaration at the top of `polls.js` for all the template-injected variables (`poll_questions`, `poll_choices`, `submitted_question_data`, plus `CKEDITOR` and `_`). Codacy was flagging `submitted_question_data` as an undeclared global since it's defined in the HTML template rather than in the JS file itself. The existing ones were pre-existing issues, so I folded them in at the same time.

Also ticked the two verification items CI can confirm — the regression tests are in the diff and the test suite passes.

CrepuscularIRIS:
Closing this PR — no maintainer activity in 2+ weeks. Happy to reopen or resubmit if you'd like to revisit. Thanks for the project!

PR #1891, opened 2026-05-02
fix(polls): fix poll answer reset on error

### LINKED ISSUE #1843: Don't make polls questions reset when form error occurs

deR0R0:
## Proposed changes
- (for multiple choice qs) Restore user's submitted answers rather than restoring from db
- Show an error message on the questions they had an error with

Currently, when the user makes an error on a poll, (For example, selecting more than the amount of choices required on a multiple choice) the user's previous poll answers were reset. This PR makes it so the user's choices are returned to them and not reset based on DB state. It also marks which question the user had an issue on :)

Closes #1843


deR0R0:
closed due to wrong thing fixed

PR #1896, opened 2026-05-08
feat(users): add user archiving feature and update tests

### LINKED ISSUE #1815: User archival feature
## Feature description
Add a feature that allows you to archive Ion users

We already archive users on IPA 
Prevents some model relationships from cascading (deleting), which can be useful
@alanzhu0 

- Set `is_active = False` and prevent archived users from showing up in search
- Add a Django admin action that allows you to do this and a management command that takes a csv of users as an input. Base this off of existing Ion management commands: [example](https://github.com/tjcsl/ion/blob/2f07af36dc895315b9b6cad1341f3a032a3e3b60/intranet/apps/dataimport/management/commands/import_users.py#L4) (put this management command in the `users` app)

Pwngo:
## Proposed changes
- Add user archiving as an admin action
- Add a management command that takes a CSV of users as input
- Add a parameter to [this file](https://github.com/tjcsl/ion/blob/2f07af36dc895315b9b6cad1341f3a032a3e3b60/intranet/apps/dataimport/management/commands/year_cleanup.py) that will archive users instead of deleting them

Closes #1815


aarushtools (changes requested):
Seems like something happened with your Git so you have a lot of unnecessary changes. Can you fix it and then I'll take a look. Thanks

PR #1899, opened 2026-05-12
fix(enrichment): default to next week on weekends

ZamanZahid:
## Proposed changes
- Added a helper function for the default enrichment date
- Updated enrichment_ context to use the new logic
- Show next week's enrichments on weekends if there are no enrichments left that weekend

Right now, the enrichment page shows the previous week on Saturdays and Sundays even though signups are closed after Friday. This changde makes it show the next week instead unless there is something happeningg that weekend.

### DIFF
--- a/intranet/apps/enrichment/views.py
+++ b/intranet/apps/enrichment/views.py
@@ -36,11 +36,29 @@ def is_weekday(date):
     return date.isoweekday() in range(1, 6)
 
 
-def enrichment_context(request, date=None):
+def get_default_enrichment_date(request):
+    """Returns next week's date on weekends unless enrichments is still happening that weekend."""
     local_time = timezone.localtime()
 
+    if is_weekday(local_time):
+        return local_time
+
+    days_until_next_week = timedelta(days=8 - local_time.isoweekday())
+    next_week_date = local_time + days_until_next_week
+
+    weekend_enrichments = EnrichmentActivity.objects.visible_to_user(request.user).filter(
+        time__range=[local_time, next_week_date],
+    )
+
+    if weekend_enrichments.exists():
+        return local_time
+
+    return next_week_date
+
+
+def enrichment_context(request, date=None):
     if date is None:
-        date = local_time
+        date = get_default_enrichment_date(request)
 
     date_today = date.replace(hour=0, minute=0, second=0, microsecond=0)
 


intranet/apps/enrichment/views.py
```diff
@@ -36,11 +36,33 @@ def is_weekday(date):
...
+def get_default_enrichment_date(request):
     local_time = timezone.localtime()
 
+    if is_weekday(local_time):
+        return local_time
+
+    weekend_start = local_time.replace(hour=0, minute=0, second=0, microsecond=0)
+
+    if timezone.is_naive(weekend_start):
+        weekend_start = timezone.make_aware(weekend_start)
```
aarushtools: why is this necessary? aren't we using django utils timezone. I could be wrong

intranet/apps/enrichment/views.py
```diff
@@ -36,11 +36,33 @@ def is_weekday(date):
...
+
+    weekend_end = weekend_start + timedelta(days=(8 - local_time.isoweekday()))
+
+    weekend_enrichments = EnrichmentActivity.objects.visible_to_user(request.user).filter(
+        time__gte=weekend_start,
+        time__lt=weekend_end,
+    )
```
aarushtools, line 51: A nitpick but I prefer time__range for consistency - it's what is already being used throughout the files (unless you don't want inclusive bounds)

intranet/apps/enrichment/views.py
aarushtools: this variable is also misleading in name - it's only the weekend start when the day is the first day of the weekend right?

intranet/apps/enrichment/views.py
```diff
@@ -36,11 +36,33 @@ def is_weekday(date):
...
+
+    weekend_start = local_time.replace(hour=0, minute=0, second=0, microsecond=0)
+
+    if timezone.is_naive(weekend_start):
+        weekend_start = timezone.make_aware(weekend_start)
+
+    weekend_end = weekend_start + timedelta(days=(8 - local_time.isoweekday()))
+
+    weekend_enrichments = EnrichmentActivity.objects.visible_to_user(request.user).filter(
+        time__gte=weekend_start,
```
aarushtools: why not just use local time + timedelta for weekend_start, why is it starting from midnight?

intranet/apps/enrichment/views.py
```diff
@@ -36,11 +36,33 @@ def is_weekday(date):
...
+        time__lt=weekend_end,
+    )
+
+    if weekend_enrichments.exists():
+        return local_time
+
+    return local_time + timedelta(days=(8 - local_time.isoweekday()))
```
aarushtools: save the timedelta as a variable that is well named so it's clear what it's doing, esp. because you repeat this line twice

intranet/apps/enrichment/views.py
```diff
@@ -36,11 +36,33 @@ def is_weekday(date):
     return date.isoweekday() in range(1, 6)
 
 
-def enrichment_context(request, date=None):
+def get_default_enrichment_date(request):
```
aarushtools, line 39: docstrings - what is this function doing?

aarushtools (changes requested):
Hi Zaman,

Thank you for the PR, can you look at my comments and implement them? Let me know if you have any questions

intranet/apps/enrichment/views.py
ZamanZahid: I added a it but,  this returns next week's date on weekends unless enrichments still are happening the same weekend.

intranet/apps/enrichment/views.py
ZamanZahid: Your right, I removed the unnecessary timezone check, and also renamed the variable to make it more clear for what it is.

intranet/apps/enrichment/views.py
ZamanZahid: Alright I Changed this to use time__range for consistency.

intranet/apps/enrichment/views.py
ZamanZahid: Saved the timedelta in a variable so it is not repeated.

intranet/apps/enrichment/views.py
ZamanZahid: Changed it so weekend_start uses local_time + timedelta rather than starting from midnight.

PR #1901, opened 2026-05-12
feat(users): add user archiving feature and update tests

### LINKED ISSUE #1815: User archival feature

Pwngo:
## Proposed changes
- Add user archiving as an admin action
- Add a management command that takes a CSV of users as input
- Add a parameter to [this file](https://github.com/tjcsl/ion/blob/2f07af36dc895315b9b6cad1341f3a032a3e3b60/intranet/apps/dataimport/management/commands/year_cleanup.py) that will archive users instead of deleting them

Closes #1815

### DIFF
--- a/intranet/apps/dataimport/management/commands/year_cleanup.py
+++ b/intranet/apps/dataimport/management/commands/year_cleanup.py
@@ -22,6 +22,10 @@ class Command(BaseCommand):
             default=get_senior_graduation_year(),
             help="The senior graduation year",
         )
+        user_mode = parser.add_mutually_exclusive_group(required=True)
+        user_mode.add_argument("--delete-users", action="store_true", dest="delete_users", help="Delete graduated users.")
+        user_mode.add_argument("--archive-users", action="store_true", dest="archive_users", help="Archive graduated users instead of deleting.")
+        user_mode.add_argument("--no-user-changes", action="store_true", dest="no_user_changes", help="Do not change users.")
 
     def ask(self, q):
         if input(f"{q} [Yy]: ").lower() != "y":
@@ -81,9 +85,16 @@ class Command(BaseCommand):
         if do_run:
             self.update_welcome()
 
-        self.stdout.write("Deleting graduated users")
-        if do_run:
-            self.handle_delete(senior_grad_year=senior_grad_year)
+        if options["delete_users"]:
+            self.stdout.write("Deleting graduated users")
+            if do_run:
+                self.handle_delete(senior_grad_year=senior_grad_year)
+        elif options["archive_users"]:
+            self.stdout.write("Archiving graduated users")
+            if do_run:
+                self.handle_archive(senior_grad_year=senior_grad_year)
+        else:
+            self.stdout.write("Skipping user delete/archive")
 
         self.stdout.write("Archiving admin comments")
         if do_run:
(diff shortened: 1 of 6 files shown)


intranet/apps/dataimport/management/commands/year_cleanup.py
```diff
@@ -112,3 +123,10 @@ def handle_delete(self, *, senior_grad_year: int):
                 usr.user_type = "alum"
                 usr.save()
                 self.stdout.write(f"User {usr.username} KEEP")
+
+    def handle_archive(self, *, senior_grad_year: int):
+        users = get_user_model().objects.filter(graduation_year__lt=senior_grad_year).exclude(user_type="alum")
+        archived_count, already_archived_count = get_user_model().archive_users(users)
```
aarushtools: I think this should be a queryset method instead of a manager method, because the input is always a filtered collection

intranet/apps/users/management/commands/archive_users.py
```diff
@@ -0,0 +1,107 @@
...
+            if not self.ask(
+                "This script will archive users (lock accounts). Ensure\nthat you know what you're doing before proceeding.\n\nContinue?"
+            ):
+                self.stdout.write("Aborted.")
+                return
+
+        archived_count, already_archived_count = user_model.archive_users(users)
+        self.stdout.write(self.style.SUCCESS(f"Archived users: {archived_count}"))
+        if already_archived_count:
+            self.stdout.write(self.style.WARNING(f"Already archived users: {already_archived_count}"))
```
aarushtools, line 86: I like what you're doing with the identifiers for filtering being derived from the CSV columns, but I think it's better if you hardcode it in to prevent anything bad from accidentally happening. Use `username` as the only column in the spreadsheet

intranet/apps/users/admin.py
```diff
@@ -1,10 +1,19 @@
-from django.contrib import admin
+from django.contrib import admin, messages
 
 from ..users.models import Course, Section, User, UserProperties
 
 
 @admin.register(User)
 class UserAdmin(admin.ModelAdmin):
+    @admin.action(description="Archive selected users")
```
aarushtools, line 8: Nice!

intranet/apps/users/models.py
```diff
@@ -842,6 +842,26 @@ def is_active(self) -> bool:
 
         return not self.username.startswith("INVALID_USER") and not self.user_locked
 
+    @classmethod
```
aarushtools: docstrings (although it will be in a different place since I don't want this on the manager)

aarushtools (changes requested):
Hi Ethan,

Thanks for the PR, you're off to a good start. In addition to my comments, does this PR do other things like removing archived users from search results? Remember it should function basically like a delete but without _actually_ deleting the user.

Pwngo:
Archiving sets user_locked=True on users, and search results are filtered to user.is_active only. is_active returns False when user_locked=True, so archived users are excluded from search results.

PR #1903, opened 2026-05-14
chore: added context to log statements

### LINKED ISSUE #729: Add additonal context to logging statements.
There are many spots in the codebase where we print out straight objects.  Often times, it is without context, like a QuerySet or a single object.  We should add context to these statements and remove unneeded statements.

This would be a good issue for someone wishing to get started with Ion development.

2028bxi:
## Proposed changes
- Added more context to logging statements
- Removed unnecessary log statements

- Makes logging more readable and clearer
Closes #729


2028bxi:
Mythofs is also my account

PR #1908, opened 2026-05-17
fix: poll question reset on validation error

### LINKED ISSUE #1843: Don't make polls questions reset when form error occurs

deR0R0:
## Proposed changes
- fixed poll question reset on validation error

originally, poll questions would reset when fields were missing. this pr fixes that by returning the submitted question data to the client and the client parsing that question data.

Closes #1843


deR0R0:
why did case fail when there was previously snake_case in the file?!?

PR #1909, opened 2026-05-17
Closes #1813

### LINKED ISSUE #1813: Scheduling enrichment activities
## Feature description
- Make a feature that allows scheduling enrichment activities over several days/weeks at once

- I meant to do this when I made the app but never got around to it
- Manually adding an enrichment activity each day gets very annoying (which is what the librarians are currently doing)
- May make the feature used more widely for other activities instead of just the library

- Possibly look at 8th period scheduling

crushedflea8060:
## Proposed changes
- Add a bulk-signup view, form, and HTML file.
- Change home.html to include the new URL.
- Mimic the original add enrichment activity feature in styling.
- New time field as well as a field that checks the current week.
## Checks done
- It has been tested in an environment.


crushedflea8060:
It seems I need to fix the commit messages and some whitespace, but I don't know how I can really change those commit messages.

PR #1911, opened 2026-05-18
feat(enrichment): implement making multiple enrichments

### LINKED ISSUE #1813: Scheduling enrichment activities

crushedflea8060:
## Proposed changes
- Add a bulk-signup view, form, and HTML file.
- Change home.html to include the new URL.
- Mimic the original add enrichment activity feature in styling.
- New time field as well as a field that checks the current week.
## Checks done
- It has been tested in an environment.
Closes #1813

### DIFF
--- a/intranet/apps/enrichment/forms.py
+++ b/intranet/apps/enrichment/forms.py
@@ -1,5 +1,9 @@
+import datetime
+from typing import Any
+
 from django import forms
 
+from ...utils.html import safe_html
 from .models import EnrichmentActivity
 
 
@@ -13,3 +17,72 @@ class EnrichmentActivityForm(forms.ModelForm):
         model = EnrichmentActivity
 
         fields = ["title", "description", "time", "location", "capacity", "presign", "groups_allowed", "groups_blacklisted"]
+
+
+WEEKDAY_FIELDS = [
+    ("monday", "Monday"),
+    ("tuesday", "Tuesday"),
+    ("wednesday", "Wednesday"),
+    ("thursday", "Thursday"),
+    ("friday", "Friday"),
+]
+
+
+class EnrichmentActivityBulkForm(EnrichmentActivityForm):
+    # A way to create multiple enrichment activities at once.
+
+    week_of = forms.DateField(
+        label="Target Week",
+        widget=forms.DateInput(attrs={"class": "datetimepicker-date"}),
+    )
+    activity_time = forms.TimeField(
+        label="Time of day",
+        widget=forms.TimeInput(attrs={"type": "time"}),
+        initial=datetime.time(12, 0),
+    )
+    monday = forms.BooleanField(label="Monday", required=False)
+    tuesday = forms.BooleanField(label="Tuesday", required=False)
+    wednesday = forms.BooleanField(label="Wednesday", required=False)
+    thursday = forms.BooleanField(label="Thursday", required=False)
+    friday = forms.BooleanField(label="Friday", required=False)
+
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+        # remove time so it doesn't appear in the form
+        self.fields.pop("time")
+
+    def clean(self) -> dict[str, Any]:
+        cleaned = super().clean()
+        # Require at least one day to be checked
+        days_checked = any(cleaned.get(day) for day, _ in WEEKDAY_FIELDS)
+        if not days_checked:
+            raise forms.ValidationError("Please select at least one day of the week.")
+        return cleaned
+
+    def get_selected_dates(self) -> list[datetime.datetime]:
+        week_of = self.cleaned_data["week_of"]
+        activity_time = self.cleaned_data["activity_time"]
+        monday = week_of - datetime.timedelta(days=week_of.weekday())
+        offsets = {
+            "monday": 0,
+            "tuesday": 1,
+            "wednesday": 2,
+            "thursday": 3,
+            "friday": 4,
+        }
+        return [
+            datetime.datetime.combine(monday + datetime.timedelta(days=offset), activity_time)
+            for day_name, offset in offsets.items()
+            if self.cleaned_data.get(day_name)
+        ]
+
+    def create_for_date(self, dt: datetime.datetime, user) -> EnrichmentActivity:
+        """Create one EnrichmentActivity for the given datetime, using this form's cleaned data."""
+        activity = self.save(commit=False)
+        activity.pk = None  # Reset the primary key because we're using one form to make multiple enrichments.
+        activity.user = user
+        activity.time = dt
+        activity.description = safe_html(activity.description)  # sanitize the html to prevent injection
+        activity.save()
+        self.save_m2m()
+        return activity
--- a/intranet/apps/enrichment/views.py
+++ b/intranet/apps/enrichment/views.py
@@ -5,18 +5,19 @@ from dateutil.relativedelta import relativedelta
 from django import http
 from django.contrib import messages
 from django.contrib.auth.decorators import login_required
+from django.db import IntegrityError, transaction
 from django.shortcuts import get_object_or_404, redirect, render
 from django.utils import timezone
 
 from ...utils.html import safe_html
 from ..auth.decorators import deny_restricted
-from .forms import EnrichmentActivityForm
+from .forms import EnrichmentActivityBulkForm, EnrichmentActivityForm
 from .models import EnrichmentActivity
 
 logger = logging.getLogger(__name__)
 
 
-def date_format(date):
+def date_format(date: datetime) -> str | None:
     try:
         d = date.strftime("%Y-%m-%d")
     except ValueError:
@@ -24,7 +25,7 @@ def date_format(date):
     return d
 
 
-def decode_date(date):
+def decode_date(date: str) -> datetime | None:
     try:
         d = datetime.strptime(date, "%Y-%m-%d")
     except ValueError:
@@ -32,11 +33,11 @@ def decode_date(date):
     return d
 
 
-def is_weekday(date):
+def is_weekday(date: datetime) -> bool:
     return date.isoweekday() in range(1, 6)
 
 
-def enrichment_context(request, date=None):
+def enrichment_context(request: http.HttpRequest, date: datetime | None = None) -> dict:
     local_time = timezone.localtime()
 
     if date is None:
@@ -72,7 +73,7 @@ def enrichment_context(request, date=None):
     return data
 
 
-def week_data(request, date=None):
+def week_data(request: http.HttpRequest, date: datetime | None = None) -> dict:
     if date:
         start_date = date
     elif "date" in request.GET:
@@ -101,7 +102,7 @@ def week_data(request, date=None):
     return data
 
 
-def month_data(request):
+def month_data(request: http.HttpRequest) -> dict:
     if "date" in request.GET:
         start_date = decode_date(request.GET["date"])
     else:
@@ -127,7 +128,7 @@ def month_data(request):
 
 @login_required
 @deny_restricted
-def enrichment_view(request):
+def enrichment_view(request: http.HttpRequest) -> http.HttpResponse:
     """Enrichment homepage.
 
     Shows a list of enrichments occurring in the next week, month, and
@@ -203,7 +204,7 @@ def enrichment_view(request):
 
 @login_required
 @deny_restricted
-def enrichment_signup_view(request, enrichment_id):
+def enrichment_signup_view(request: http.HttpRequest, enrichment_id: int) -> http.HttpResponse:
     if request.method == "POST":
         enrichment = get_object_or_404(EnrichmentActivity, id=enrichment_id)
 
@@ -240,7 +241,7 @@ def enrichment_signup_view(request, enrichment_id):
 
 @login_required
 @deny_restricted
-def enrichment_roster_view(request, enrichment_id):
+def enrichment_roster_view(request: http.HttpRequest, enrichment_id: int) -> http.HttpResponse:
     is_enrichment_admin = request.user.has_admin_permission("enrichment")
     enrichment = get_object_or_404(EnrichmentActivity, id=enrichment_id)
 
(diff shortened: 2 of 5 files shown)


crushedflea8060:
I was informed by Bryce that Codacy doesn't matter and that it will resolve in the merge.

crushedflea8060:
I implemented safe_html today as well as saving the object user because I forgot to.

crushedflea8060:
oh another note, i had to manually modify the form because there were two instances of the checkboxes, let me know if there's a better fix.

makors:
@greptileai review

crushedflea8060:
@greptileai review

intranet/apps/enrichment/forms.py
```diff
@@ -1,3 +1,5 @@
+import datetime  # for mass form
```
aarushtools, line 1: unnecessary comment

intranet/apps/enrichment/forms.py
```diff
@@ -13,3 +15,57 @@ class Meta:
...
+
+    def clean(self):
+        cleaned = super().clean()
+        # Require at least one day to be checked
+        days_checked = any(cleaned.get(day) for day, _ in WEEKDAY_FIELDS)
+        if not days_checked:
+            raise forms.ValidationError("Please select at least one day of the week.")
+        return cleaned
+
+    def get_selected_dates(self):  # returns in the form of a list.
```
aarushtools, line 56: type annotate the return instead of a comment

intranet/apps/enrichment/forms.py
```diff
@@ -13,3 +15,57 @@ class Meta:
...
+    thursday = forms.BooleanField(label="Thursday", required=False)
+    friday = forms.BooleanField(label="Friday", required=False)
+    activity_time = forms.TimeField(label="Time of day", widget=forms.TimeInput(attrs={"type": "time"}), initial=datetime.time(12, 0))
+
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+        # remove time so it doesn't appear in the form
+        self.fields.pop("time")
+
+    def clean(self):
```
aarushtools, line 48: type annotate

intranet/apps/enrichment/views.py
```diff
@@ -309,13 +310,62 @@ def add_enrichment_view(request):
     return render(request, "enrichment/add_modify.html", context)
 
 
+@login_required
+@deny_restricted
+def enrichment_bulk_create_view(request):
```
aarushtools, line 315: type everything

intranet/apps/enrichment/views.py
```diff
@@ -309,13 +310,62 @@ def add_enrichment_view(request):
...
+    """View for creating multiple enrichments.
+
+    Get requests will render bulk-create form with Mon-Fri checkboxes.
+    Post requests create one EnrichmentActivity for each checked day and then redirect.
+    """
+    is_enrichment_admin = request.user.has_admin_permission("enrichment")
+    if not is_enrichment_admin:
+        raise http.Http404
```
aarushtools, line 323: messages.error and redirect instead - see existing implementations of this permission denial in other parts of ion for ideas

intranet/apps/enrichment/views.py
```diff
@@ -309,13 +310,62 @@ def add_enrichment_view(request):
...
+
+                    for dt in form.get_selected_dates():
+                        activity = form.save(commit=False)
+                        activity.pk = None
+                        activity.user = request.user
+                        activity.time = dt
+                        activity.description = safe_html(activity.description)  # sanitize the html to prevent injection
+                        activity.save()
+                        form.save_m2m()
+                        created.append(dt.strftime("%A, %b %-d"))
```
aarushtools, line 340: This is quite confusing and a bit hacky, why not just call .create()?

intranet/apps/enrichment/views.py
```diff
@@ -309,13 +310,62 @@ def add_enrichment_view(request):
...
+
+                messages.success(
+                    request,
+                    f"Created {len(created)} enrichment activities: " + ", ".join(created),
+                )
+                return redirect("enrichment")
+            except IntegrityError:
+                form.add_error(None, "Encountered an error in adding activities, maybe one of the dates already has one scheduled?")
```
aarushtools, line 348: This is a bit broad, is it possible to list the exact dates?

intranet/apps/enrichment/views.py
```diff
@@ -309,13 +310,62 @@ def add_enrichment_view(request):
...
+
+
 @login_required
 @deny_restricted
 def modify_enrichment_view(request, enrichment_id):
-    """Modify enrichment activity page.
+    """
+    Modify enrichment activity page.
 
     Args:
```
aarushtools, line 366: type annotate entire function while you're at it

intranet/templates/enrichment/bulk_create.html
```diff
@@ -0,0 +1,74 @@
+{% extends "page_with_nav.html" %}
```
aarushtools, line 1: For this entire file, always indent in a django block. For example in main you kept the html flush, though it's correct for script

intranet/templates/enrichment/bulk_create.html
```diff
@@ -0,0 +1,74 @@
...
+        {% stylesheet 'dark/nav' %}
+        {% stylesheet 'dark/events' %}
+    {% endif %}
+{% endblock %}
+{% block main %}
+<div class="events enrichment announcements primary-content">
+    <h2>Bulk Create Enrichment Activities</h2>
+    <p>Choose a week for your activity, and check off the days you want. One activity will be created for each checked day.</p>
+    <form method="post" id="event_form">
+    <table>
```
aarushtools, line 40: Indent correctly

intranet/templates/enrichment/bulk_create.html
```diff
@@ -0,0 +1,74 @@
...
+    {% csrf_token %}
```
aarushtools, line 41: Should be under form not table, I think

intranet/templates/enrichment/bulk_create.html
```diff
@@ -0,0 +1,74 @@
...
+    <tr><th><label for="id_title">Title:</label></th><td>{{ form.title }}{{ form.title.errors }}</td></tr>
+    <tr><th><label for="id_description">Description:</label></th><td>{{ form.description }}{{ form.description.errors }}</td></tr>
+    <tr><th><label for="id_capacity">Capacity:</label></th><td>{{ form.capacity }}{{ form.capacity.errors }}</td></tr>
+    <tr><th><label for="id_location">Location:</label></th><td>{{ form.location }}{{ form.location.errors }}</td></tr>
+    <tr><th><label for="id_groups_allowed">Groups allowed:</label></th><td>{{ form.groups_allowed }}{{ form.groups_allowed.errors }}</td></tr>
+    <tr><th><label for="id_groups_blacklisted">Groups blacklisted:</label></th><td>{{ form.groups_blacklisted }}{{ form.groups_blacklisted.errors }}</td></tr>
+    <tr><th><label for="id_presign">Presign:</label></th><td>{{ form.presign }}{{ form.presign.errors }}</td></tr>
+    <tr><th><label for="id_week_of">Week of:</label></th><td>{{ form.week_of }}{{ form.week_of.errors }}</td></tr>
+    <tr><th><label for="id_activity_time">Time of day:</label></th><td>{{ form.activity_time }}{{ form.activity_time.errors }}</td></tr>
```
aarushtools, line 50: I would prefer if you indented these a little bit, it's not readable easily

intranet/templates/enrichment/bulk_create.html
```diff
@@ -0,0 +1,74 @@
...
+                {# creates a day checkbox within the weekday checkboxes for each day #}
+                <label class="day-checkbox">{{ form.monday }} {{ form.monday.label }}</label>
+                <label class="day-checkbox">{{ form.tuesday }} {{ form.tuesday.label }}</label>
+                <label class="day-checkbox">{{ form.wednesday }} {{ form.wednesday.label }}</label>
+                <label class="day-checkbox">{{ form.thursday }} {{ form.thursday.label }}</label>
+                <label class="day-checkbox">{{ form.friday }} {{ form.friday.label }}</label>
+            </div>
+            {% if form.non_field_errors %}
+                <ul class="errorlist">{{ form.non_field_errors }}</ul>
+            {% endif %}
```
aarushtools, line 64: Put this at the top of the form, or else it looks like a field error for days of the week

aarushtools: Also, doesn't form.non_field_errors already wrap it in html?

intranet/templates/enrichment/bulk_create.html
aarushtools: Also can't you define the label in django and then use {{ form.field.label_tag }}

intranet/templates/enrichment/bulk_create.html
```diff
@@ -0,0 +1,74 @@
+{% extends "page_with_nav.html" %}
+{% load static %}
+{% load dates %}
```
aarushtools, line 3: Is this being used (load dates)

intranet/templates/enrichment/bulk_create.html
```diff
@@ -0,0 +1,74 @@
...
+    {{ block.super }}
+    <script src="{% static 'vendor/selectize.js-0.12.4/dist/js/standalone/selectize.min.js' %}"></script>
+    <script>
+        $(function() {
+            $("#id_groups_allowed").selectize({
+                create: false
+            })
+            $("#id_groups_blacklisted").selectize({
+                create: false
+            });
```
aarushtools, line 18: Can you do 
```js
$(function() {
    $("#id_groups_allowed, #id_groups_blacklisted").selectize({
        create: false
    });
});
```
or does it not work

intranet/templates/enrichment/home.html
```diff
@@ -101,6 +101,7 @@ <h2>Enrichment</h2>
                     <i class="fas fa-plus"></i>
                     Add
                 </a>
+                <a href="{% url 'enrichment_bulk_create' %}" class="button"> <i class="fas fa-plus"></i> Bulk Create (by week) </a>
```
aarushtools, line 104: just "bulk create"
(That text is too long for a button)

intranet/templates/enrichment/bulk_create.html
aarushtools, line 50: Use datetimepicker extension (see what the normal add view uses, copy that)

intranet/templates/enrichment/bulk_create.html
```diff
@@ -0,0 +1,74 @@
...
+                <label class="day-checkbox">{{ form.friday }} {{ form.friday.label }}</label>
+            </div>
+            {% if form.non_field_errors %}
+                <ul class="errorlist">{{ form.non_field_errors }}</ul>
+            {% endif %}
+        </td>
+    </tr>
+    <tr><td>&nbsp;</td><td>
+        <input type="submit" style="width: 270px" value="Bulk Create Enrichment">
+        <a href="{% url 'enrichment' %}" class="button">Cancel</a>
```
aarushtools, line 69: Follow existing behavior; the add screen does not have cancel (you can press back arrow), so it seems unnecessary here

aarushtools (changes requested):
Hi Finnegan,

Good start on your PR, please address the comments. This is just at a first glance so I might be wrong for some things, if so you can ignore it. Thanks!

intranet/templates/enrichment/bulk_create.html
crushedflea8060: I copied that from add_modify.html for the sake of consistency. I'm not sure if either of them use it though.

intranet/templates/enrichment/bulk_create.html
crushedflea8060: in the original file it's under `<table>`

intranet/apps/enrichment/views.py
crushedflea8060: can you refer me to a place where there's an implementation of this? I've been looking for about 15 minutes and I haven't found it yet

aarushtools: https://github.com/tjcsl/ion/blob/e7976cefb330056852e1586619f92a991c0bd99f/intranet/apps/auth/decorators.py#L48
Just search messages.error

crushedflea8060: i found an implementation in `cslapps/views.py`. I'm redirecting to `/`, is that fine?

intranet/apps/enrichment/views.py
crushedflea8060: wouldn't that be even more complex? If you need I can add comments, but I'm pretty sure it's fine right now. (i consulted claude for this and it said that it would be a step down to call create instead.)

intranet/apps/enrichment/views.py
crushedflea8060: As far as I know we can't really track which dates failed given it rolls back the entire thing. It might be better to just have an error that causes the entire thing to fail just so there's more care put into the process of making the activities.

intranet/templates/enrichment/bulk_create.html
crushedflea8060: yes, my bad

intranet/apps/enrichment/views.py
aarushtools: you can just check beforehand though, instead of catching an integrityerror which may not necessarily be date-related

intranet/apps/enrichment/views.py
aarushtools: The main issue I have with this is doing the activity.pk = None trick. How would calling .create be more complex? It doesn't necessarily need to be that either, you could also attach the creation to a method on the form (perhaps better semantically)

crushedflea8060: yeah i can do a method, although I think the pk thing is kind of necessary

intranet/apps/enrichment/views.py
```diff
@@ -311,11 +310,55 @@ def add_enrichment_view(request):
...
+    if not is_enrichment_admin:
+        messages.error(request, "You are not authorized to access the requested app.")
+        return redirect("/")
+
+    if request.method == "POST":
+        form = EnrichmentActivityBulkForm(data=request.POST)
+        if form.is_valid():
+            current_dt = None
+            try:
+                with transaction.atomic():
```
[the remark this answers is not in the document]
crushedflea8060: greptile failed here i believe.

PR #1919, opened 2026-05-24
Add watermark to first page in print jobs (bottom right corner)

### LINKED ISSUE #1879: Minimize paper used across jobs
## Feature description
- Find a way to minimize paper used when printing job names. Currently, whenever someone submits a new print job to one of the Ion printers, a whole separate page with their username and job info is printed as well as the actual job itself. Over time, this adds up to a lot of waste as the separate page is useless otherwise.

- Paper running out is an issue with the printers, and minimizing unneeded printing is environmentally and waste friendly

- You need to think about several different implementations and weigh the pros and cons
- Ideally, a watermark marking the user's job would be printed in a small spot somewhere on the user's paper, so we don't need to print a whole separate paper to mark the user's job. How will you choose a blank spot? What will the watermark look like? It needs to be noticeable and separable, and if the spot is dynamic (changing) the user should know where it will land before printing the job. Another method is to shrink the page and add the watermark at the bottom of the page, but it has low visibility among other cons. Both approaches have cons. You may have another approach

blues-hue:
## Proposed changes
- Adds a watermark to the first page on the bottom right corner on print jobs (Ion user and job id)
- Some risks are that the system is only created so far, so the user can't really see where the watermark will go, but the location is hardcoded based on page width
- Deleted the title page creation

- Reduce paper waste from creating a whole separate page of a watermark
Closes #1879

### DIFF
--- a/intranet/apps/printing/views.py
+++ b/intranet/apps/printing/views.py
@@ -345,7 +345,7 @@ def add_watermark(pdf_path: str, obj) -> str:
     reader = PdfReader(pdf_path)
     writer = PdfWriter()
 
-    out_fd, out_path = tempfile.mkstemp(suffix="_wm.pdf")
+    out_path = tempfile.mkstemp(suffix="_wm.pdf")
     i = 0
 
     for page in reader.pages:
--- a/intranet/apps/printing/views.py
+++ b/intranet/apps/printing/views.py
@@ -16,7 +16,6 @@ from django.core.cache import cache
 from django.shortcuts import redirect, render
 from django.template.loader import get_template
 from django.utils.text import slugify
-from django.utils import timezone
 from pypdf import PdfReader, PdfWriter
 from reportlab.pdfgen import canvas
 from sentry_sdk import add_breadcrumb, capture_exception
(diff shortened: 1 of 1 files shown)


aarushtools:
No difference from existing impl. #1883 
You can reopen it if you have a different solution

blues-hue:
> No difference from existing impl. #1883 You can reopen it if you have a different solution

@aarushtools sorry, what would you say is a different implementation? Would you consider shrinking the page to about 98% of its original size then adding the watermark at the bottom as a different implementation? Or should I have an entirely different solution (i.e. i thought of adding a qr code linking to a page with their details)

crushedflea8060:
@blues-hue just my input, but I wouldn't recommend doing a QR code since we won't _technically_ be allowed to use phones next year. Also, what's your rationale for making an alternative PR? There's another PR that closes this same issue, as Aarush said.

blues-hue:
> @blues-hue just my input, but I wouldn't recommend doing a QR code since we won't _technically_ be allowed to use phones next year. Also, what's your rationale for making an alternative PR? There's another PR that closes this same issue, as Aarush said.

@crushedflea8060 You're right about the QR code, I did not think about that. My rationale for an alternative PR is **only** making the watermark on the first page, as having a watermark on every page could be kind of annoying in some print jobs. Also, I originally made this alternative PR for the Understudy Program (Aarush said I could do an alternative implementation).

PR #1926, opened 2026-07-21
feat: Add code-based attendance option

### LINKED ISSUE #1612: Alternate attendance options

TigerCodeHub:
Closes #1612

> **Note:** This supersedes #1770. The work was moved from the `dev` branch to a dedicated `attendance` branch to keep the fork's `dev` clean and in sync with upstream. See #1770 for prior review history.

## Proposed changes
- Added a new method of attendance. For each EighthScheduledActivity, a random 6-character code 0-9 & A-Z is generated. This is displayed to the teacher (and eighth admins). Teachers can share the code to students at the activity, who can then input the code on a new eighth/student_attendance page which marks them as present if the code is correct (and shows a message as such).
- This also sets the attendance_taken on a sch. activity to True (as it would be odd otherwise).
- Note: Migration is required (new attendance_code attribute for EightScheduledActivity object).

- Attendance will only be marked if the code is given during the block (2:25 - 3:05 for A blocks, 3:15 - 4:00 for B blocks, anytime for other kinds of blocks). Otherwise, the student will get an error message.

- Students may access the attendance page through buttons which will appear on the eighth/glance (location.html) and eight/signup (signup.html) pages from 2:25 to 4:00.
- This attendance method coexists with already-existing methods (so the last change is the one which has effect).
- (All above times in PM).

This method (credits to Alan Zhu for the idea) will let students mark themselves present if they are at an activity (and thus receive the code from the teacher). This makes it more convenient for the teacher to take attendance, as all they need to do is show the code to the students at the activity.


makors:
@greptileai review

ElijahFeldman7:
Can we revisit this? I think it would be really useful

PR #1929, opened 2026-08-05
feat(user): add administrator field (based on counselor) for students

TigerCodeHub:
## Proposed changes
- Adds an "administrator" field to students and counselors which links to their subschool administrator's Ion account, similar to how the current student counselor setup works.
- Incorporated this into HTML templates and import student data commands (along with a new command update_administrators to update this for current students).
- Link to list of administrators, counselors, etc. - implemented via a "Subschools" button on the dashboard.
- Closes #1861 in the process by adding dropdowns for both administrator and counselor selection in editing a profile.

Requested by the administration.

### DIFF
--- a/intranet/apps/dataimport/management/commands/update_administrators.py
+++ b/intranet/apps/dataimport/management/commands/update_administrators.py
@@ -1,0 +1,129 @@
+#!/usr/bin/env python3
+
+import csv
+import sys
+from typing import Any
+
+from django.contrib.auth import get_user_model
+from django.core.management.base import BaseCommand, CommandError, CommandParser
+from django.db import transaction
+
+from intranet.utils.date import get_senior_graduation_year
+
+from ....users.models import User
+
+
+class Command(BaseCommand):
+    help = "Update subschool administrator information, inheriting from each student's counselor"
+
+    def add_arguments(self, parser: CommandParser) -> None:
+        parser.add_argument(
+            "filename",
+            type=str,
+            nargs="?",
+            default=None,
+            help="Path to SIS import CSV with a Student ID column, and optionally an Administrator column",
+        )
+
+        parser.add_argument("--all", action="store_true", dest="all", default=False, help="Update every student who has not graduated")
+        parser.add_argument(
+            "--graduation-years",
+            dest="graduation_years",
+            nargs="+",
+            type=int,
+            default=None,
+            metavar="YEAR",
+            help="Update only students graduating in these years (e.g. --graduation-years 2027 2028)",
+        )
+
+        parser.add_argument("--run", action="store_true", dest="run", default=False, help="Actually modifies the DB")
+
+    def handle(self, *args: str, **kwargs: Any) -> None:
+        filename = kwargs["filename"]
+        all_students = kwargs["all"]
+        graduation_years = kwargs["graduation_years"]
+
+        if sum(map(bool, (filename, all_students, graduation_years))) != 1:
+            raise CommandError("Pass exactly one of: a CSV filename, --all, or --graduation-years")
+
+        to_run = kwargs["run"]
+
+        if not to_run:
+            sys.stdout.write(
+                """This script is running in pretend mode.
+Pass --run to actually run this script.
+Please MAKE SURE you have updated info before running this script.
+Actually running is a destructive operation.
+"""
+            )
+
+        # Collect the students to work through, paired with any administrator named for them
+        targets = []
+        if filename:
+            try:
+                with open(filename, encoding="utf-8") as f:
+                    data = list(csv.DictReader(f))
+            except OSError as ex:
+                raise CommandError(str(ex)) from ex
+
+            if data and "Student ID" not in data[0]:
+                raise CommandError(f"{filename} has no 'Student ID' column")
+
+            for row in data:
+                sid = row["Student ID"].strip()
+                user = get_user_model().objects.user_with_student_id(sid)
+                if user is None:
+                    sys.stdout.write(f"There is no Ion account found for SID {sid}\n")
+                    continue
+
+                targets.append((user, (row.get("Administrator") or "").strip()))
+        else:
+            # Every one of these runs through the counselor chain, so pull it in up front
+            students = get_user_model().objects.filter(user_type="student").select_related("counselor__administrator")
+            if graduation_years:
+                students = students.filter(graduation_year__in=graduation_years)
+            else:
+                students = students.filter(graduation_year__gte=get_senior_graduation_year())
+
+            sys.stdout.write(f"Working through {students.count()} student(s).\n")
+            targets = [(user, "") for user in students]
+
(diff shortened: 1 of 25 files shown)


intranet/apps/dataimport/management/commands/update_administrators.py
```diff
@@ -0,0 +1,136 @@
...
+        if sum(map(bool, (filename, all_students, graduation_years))) != 1:
+            raise CommandError("Pass exactly one of: a CSV filename, --all, or --graduation-years")
+
+        to_run = kwargs["run"]
+
+        if not to_run:
+            sys.stdout.write("This script is running in pretend mode.\n")
+            sys.stdout.write("Pass --run to actually run this script.\n")
+            sys.stdout.write("Please MAKE SURE you have updated info before running this script.\n")
+            sys.stdout.write("Actually running is a destructive operation.\n")
```
aarushtools: Can we use """?

aarushtools (changes requested):
Thanks for the PR Tiger, it looks pretty close to ready (I know it's still in draft)
Two things I would note:
- Try to minimize helper functions if they are only called once - just embed them in the code (unless it's a specific case like inside a try; except that would look convoluted without it.).
- Type annotate your function definitions

I haven't had a chance to test yet if this is already implemented, can you change what administrator a counselor is assigned to from the UI as well?

TigerCodeHub:
Thanks for the feedback Aarush! 

Ironically the only way for an admin to edit a user's information is at /eighth/profile/edit/<user-id>, and the counselor field there was an IntegerField (takes the counselor's Ion DB id) - I've added something similar there to set an administrator and made them both a searchable dropdown rather than this integer ID box. 

PR update will be here shortly.

TigerCodeHub:
Added the final subschool display page at /subschools with a button on the Dashboard. The names are linked to mailto email addresses.

Subschool information is hardcoded in settings/__init__.py with Ion usernames - emails are looked up via "tj_email" (which displays the fcps email for these staff in prod) - the email for Ms. Wilson was hardcoded because she does not have a user in production Ion (and the email is public on FCPS's website anyways). 

Two screenshots are attached showing the new page.
image
image

intranet/apps/dataimport/management/commands/update_administrators.py
```diff
@@ -0,0 +1,130 @@
...
+                    sys.stdout.write(f"Switching administrator for {user.username} from {user.administrator} to {administrator}\n")
+                    if to_run:
+                        user.administrator = administrator
+                        user.save(update_fields=["administrator"])
+
+    def resolve_administrator(self, user: User, listed_name: str) -> User | None:
+        """Returns the administrator to assign ``user``, or ``None`` if it cannot be determined."""
+        # Kept out of handle() because the try/except around the lookup, plus the counselor
+        # fallback, pushes handle() past the complexity limit when inlined.
+        # We assume every administrator has a unique last name; if not, please edit this file
```
aarushtools: Probably remove "if not, please edit this file"?

intranet/apps/dataimport/management/commands/update_administrators.py
```diff
@@ -0,0 +1,130 @@
...
+        name = listed_name.split(",")[0].strip()
+        if name:
+            # Administrators have no dedicated user_type, so search every non-student account
+            try:
+                return get_user_model().objects.exclude(user_type="student").get(last_name=name)
```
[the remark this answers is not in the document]
aarushtools: Administrators are defined as `teacher` so yes you don't need to do the exclude thing here it should be a simple query

intranet/apps/users/views.py
```diff
@@ -109,6 +111,31 @@ def profile_view(request, user_id=None):
...
+
+    # Usernames are unique per case in the database, so match the configured spelling first and only
+    # fall back to a case-insensitive match. tj_email filters each user's addresses, so it costs a
+    # query per person; that is fine for a list this small.
+    lookups = [person for person in staff if not person.get("email") and person.get("username")]
+    users = get_user_model().objects.annotate(match_name=Lower("username")).filter(match_name__in=[p["username"].lower() for p in lookups])
+    by_exact = {user.username: user for user in users}
+    by_lower = {user.match_name: user for user in users}
+    for person in lookups:
+        user = by_exact.get(person["username"]) or by_lower.get(person["username"].lower())
```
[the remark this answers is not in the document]
aarushtools: Honestly this entire fallback isn't necessary at all

intranet/settings/__init__.py
```diff
@@ -106,6 +106,50 @@
 NOMINATIONS_ACTIVE = False
 NOMINATION_POSITION = ""
 
+# Subschool staff, listed at /subschools. Update as assignments change.
```
aarushtools: Looks good, but can you fake this data with known fake data accounts (see usernames.md in docs), I'll just override this in prod

intranet/templates/dashboard/dashboard.html
```diff
@@ -121,6 +121,10 @@ <h2>{{ dashboard_header }}</h2>
                 <a href="{% url 'index' %}" class="button" style="float:left"><i class="fas fa-arrow-left" style="width: 11px"></i> Dashboard</a>
                 &nbsp;
             {% endif %}
+            <a class="button subschools-button" href="{% url 'subschools' %}">
```
aarushtools: Image
can you put it under tj resources? we can move it into an actual button if admin requests later. Also default to only teachers can view (and admins)

aarushtools (changes requested):
Hi Tiger. Good work so far, very close to merging. See my comments and also:
- Your approach to the dropdown looks a bit convoluted in logic, I think we can cut out the whole id thing here. I remember Eli made a PR for this that looked a bit different, maybe reference that

intranet/settings/__init__.py
aarushtools: also can you add last updated config setting and it shows on the subschools page at the bottom as well

intranet/templates/dashboard/dashboard.html
TigerCodeHub: Sure I can move it to TJ Resources.
But why only teachers/admin view? (its not sensitive info and such a restriction doesn't seem to be requested.)

intranet/settings/__init__.py
TigerCodeHub: Done.

TigerCodeHub:
I've resolved the issues and matched the dropdown selection logic to how Eli did it in his PR, and the subschools link is now in TJ Resources. I left this commit separate for easier review, but if its good I'll squash them before merging.

intranet/templates/dashboard/dashboard.html
aarushtools: I was thinking about clutter but we can just move it later if needed

aarushtools (approved):
Thanks, looks good! Can you squash

TigerCodeHub:
Squashed and should be good to go!

